PR #10, opened 2020-07-18
Add nvm version display.

xolve:
The following test cases fails, would appreciate if you can help me out in testing these:
* pwd
* status

I have chosen a less distracting green for nvm, since nodejs's default icon color is green. Also I think default color of virtual_env item should be yellow since python is yellow IMHO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,11 @@ Tide variables that users are encouraged to modify begin with the string `tide_`
 
 ## General Variables
 
-| Variable           | Description                                | Default                                      |
-| ------------------ | ------------------------------------------ | -------------------------------------------- |
-| newline            | add empty line before each prompt          | true                                         |
-| left_prompt_items  | order of the left prompt items to display  | pwd git_prompt newline prompt_char           |
-| right_prompt_items | order of the right prompt items to display | status cmd_duration context jobs virtual_env |
+| Variable           | Description                                | Default                                          |
+| ------------------ | ------------------------------------------ | ------------------------------------------------ |
+| newline            | add empty line before each prompt          | true                                             |
+| left_prompt_items  | order of the left prompt items to display  | pwd git_prompt newline prompt_char               |
+| right_prompt_items | order of the right prompt items to display | status cmd_duration context jobs virtual_env nvm |
 
 ### prompt_connection
 
@@ -155,6 +155,13 @@ Tide's git capabilities are inherited from fish's built-in [fish_git_prompt][]. 
 | virtual_env_color   | color of virtual_env item                                                                                                              | 00AFAF      |
 | virtual_env_icon    | icon to display in front of virtual_env item                                                                                           | ''         |
 | virtual_env_display | Options are venvName and projectName. venvName is the virtual env directory, most often .venv. projectName is the directory above that | projectName |
+
+### nvm
+
+| Variable      | Description                          | Default |
+| ------------- | ------------------------------------ | ------- |
+| nvm_env_color | color of nvm item                    | 00AFAF  |
+| nvm_env_icon  | icon to display in front of ncm item | ''     |
 
 ## Fonts
 
--- a/functions/_tide_item_nvm.fish
+++ b/functions/_tide_item_nvm.fish
@@ -1,0 +1,9 @@
+function _tide_item_nvm
+    test 
+    set -l node_version (functions -q nvm; and nvm current)
+
+    if test "$node_version" != "none" -a "$node_version" != "system" -a -n "$node_version"
+        set_color $tide_nvm_color
+        printf '%s ' "$tide_nvm_icon $node_version"
+    end
+end
--- a/tide_theme/configure/configs/lean.fish
+++ b/tide_theme/configure/configs/lean.fish
@@ -10,7 +10,7 @@ _set fake_tide_prompt_connection_icon ' '
 
 # --------------------Prompt Items--------------------
 _set fake_tide_left_prompt_items 'pwd' 'git_prompt' 'newline' 'prompt_char'
-_set fake_tide_right_prompt_items 'status' 'cmd_duration' 'context' 'jobs' 'virtual_env'
+_set fake_tide_right_prompt_items 'status' 'cmd_duration' 'context' 'jobs' 'virtual_env' 'nvm'
 # ------------Prompt Char------------
 _set fake_tide_prompt_char_success_color $_tide_color_green
 _set fake_tide_prompt_char_failure_color FF0000
@@ -61,6 +61,9 @@ _set fake_tide_jobs_color $tideColorDarkerGreen
 _set fake_tide_virtual_env_color 00AFAF
 _set fake_tide_virtual_env_display 'projectName'
 _set fake_tide_virtual_env_icon ''
+# ---------------NVM---------------
+_set fake_tide_nvm_color 22BB99
+_set fake_tide_nvm_icon ''
 # ---------------Time---------------
 _set fake_tide_time_color 5F8787
 _set fake_tide_time_format '%T'
--- a/tide_theme/configure/configs/pure.fish
+++ b/tide_theme/configure/configs/pure.fish
@@ -60,6 +60,9 @@ _set fake_tide_jobs_color $tideColorDarkerGreen
 _set fake_tide_virtual_env_color 00AFAF
 _set fake_tide_virtual_env_display 'projectName'
 _set fake_tide_virtual_env_icon ''
+# ---------------NVM---------------
+_set fake_tide_nvm_color 22BB99
+_set fake_tide_nvm_icon ''
 # ---------------Time---------------
 _set fake_tide_time_color 6C6C6C
 _set fake_tide_time_format '%T'
--- a/tide_theme/configure/prompt_items/_fake_tide_item_nvm.fish
+++ b/tide_theme/configure/prompt_items/_fake_tide_item_nvm.fish
@@ -1,0 +1,2 @@
+function _fake_tide_item_nvm
+end


README.md
```diff
@@ -59,11 +59,11 @@ Tide variables that users are encouraged to modify begin with the string `tide_`
 
 ## General Variables
 
-| Variable           | Description                                | Default                                      |
-| ------------------ | ------------------------------------------ | -------------------------------------------- |
-| newline            | add empty line before each prompt          | true                                         |
-| left_prompt_items  | order of the left prompt items to display  | pwd git_prompt newline prompt_char           |
-| right_prompt_items | order of the right prompt items to display | status cmd_duration context jobs virtual_env |
+| Variable           | Description                                | Default                                          |
```
IlanCosman, line 62: I should probably have [Prettier](https://github.com/prettier/prettier) as part of the CI workflow, but could you run Prettier on this file to fix these formatting issues? You can install it as a VSCode extension or as a cli command.

functions/_tide_item_nvm.fish
```diff
@@ -0,0 +1,11 @@
+function _tide_item_nvm
+    set -l node_version (nvm current 2> /dev/null)
```
IlanCosman, line 2: unfortunately `nvm current 2>/dev/null` does not prevent fish from complaining and spamming the user with errors if an nvm command is not available. Add a `test functions -q nvm` instead of 2>/dev/null to prevent these errors.

functions/_tide_item_nvm.fish
```diff
@@ -0,0 +1,11 @@
...
+
+    if test -n "$node_version"
```
IlanCosman, line 4: The test command has the `-a` flag to combine expressions. I haven't verified it, but in general more commands = more time. Please combine these expressions with `-a`.

README.md
```diff
@@ -156,6 +156,13 @@ Tide's git capabilities are inherited from fish's built-in [fish_git_prompt][].
 | virtual_env_icon    | icon to display in front of virtual_env item                                                                                           | ''         |
 | virtual_env_display | Options are venvName and projectName. venvName is the virtual env directory, most often .venv. projectName is the directory above that | projectName |
 
+### nvm
```
IlanCosman, line 159: I should probably have [Prettier](https://github.com/prettier/prettier) as part of the CI workflow, but could you run Prettier on this file to fix this formatting issues? You can install it as a VSCode extension or as a cli command.

IlanCosman:
Hi there @xolve, thank you so much for your PR! 😄 I left a few comments on your files. 

Additionally there's the issue of using `nvm current` in fish. The most popular fish nvm solution (by stars) is <https://github.com/jorgebucaran/fish-nvm>. This nvm solution does not have an `nvm current` command. 

Other nvm solutions like <https://github.com/FabioAntunes/fish-nvm> and <https://github.com/brigand/fast-nvm-fish> use bash to get to nvm. This is less consistent, for example on my machine nvm added itself to my .zshrc, not .bashrc (which i didn't have at the time because I deleted it). 

nvm in fish is a real can of worms 😞 

Perhaps we could somehow use `node --version` and see if that differs from the known system version?

README.md
```diff
@@ -156,6 +156,13 @@ Tide's git capabilities are inherited from fish's built-in [fish_git_prompt][].
...
+
+| Variable        | Description                                                                                                                    | Default     |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------ | ----------- |
+| nvm_env_color   | color of nvm item                                                                                                              | 00AFAF      |
```
IlanCosman, line 163: These should begin with `nvm` not `nvm_env` as the item is call nvm, and the header for the section is called nvm

IlanCosman:
Also you can ignore the failing tests, I only started updating/taking them seriously in preparation for 2.0.0.

functions/_tide_item_nvm.fish
xolve: Done. The check would look like `(functions -q nvm; and nvm current)`

functions/_tide_item_nvm.fish
xolve: Also completed and tested :)

README.md
xolve: This is a pretty cool tool! And I think I will use it more often :smile: I have used it on command line with no options and I think you must be using it simialrily.

xolve:
The method I have used here is pretty straight out of the how `theme-bobthefish` does it here: https://github.com/oh-my-fish/theme-bobthefish/blob/a2ad38aa051aaed25ae3bd6129986e7f27d42d7b/fish_prompt.fish#L853

Using `node` command is quite opinionated as it can display node version from system package from distro or custom install of a user. Objective of this function is to display current `nvm` environment only.

xolve:
`nvm` is not directly supported on fish and addon packages are recommended ways to use it. Once they are installed `nvm` command being preset in PATH is valid expectation IMHO.

IlanCosman:
> `nvm` is not directly supported on fish and addon packages are recommended ways to use it. Once they are installed `nvm` command being preset in PATH is valid expectation IMHO.

I agree but how should we support the various nvm addon packages for fish? As I mentioned above the most popular one has no `nvm current` command. It seems we either have to use `node --version` to support all of them, or just pick one and tailor to that. If it came to that I'd prefer to go with Jorgebucaran's--it's the most popular and I trust it'll be maintained for a long time.

xolve:
My earlier opinion was that: You wouldn't need to worry about how users got `nvm` into their fish environment. I am using a much simpler method from here: https://eshlox.net/2019/01/27/how-to-use-nvm-with-fish-shell and it works as well.

I have tested my approach with [OMF's nvm plugin](https://github.com/derekstavis/plugin-nvm) and a very simple method in a blog post [here](https://eshlox.net/2019/01/27/how-to-use-nvm-with-fish-shell). They are similar in approach that they provide a pass through to `nvm` command.

Approach by [jorgebucaran' nvm plugin](https://github.com/jorgebucaran/fish-nvm) is to wrap around each nvm sub-command. IMHO it won't scale because you are always trying to catch-up with upstream. I wouldn't recommend it.

Going by GitHub stars first one is more popular.

However we cannot allow users to see breaking changes. I think keep `node` as version check sounds okay for now.

IlanCosman:
Alright. It sounds like you'd prefer to go with the `node --version` method. The issue is figuring out what the system node version is. 

One perhaps inelegant option is to set a `tide_node_system_version` variable when Tide is installed. The user would have to manually update this. Also, if a different node was being used at the time of installation, this method would set the incorrect node. 

Any other ideas? I'm struggling to come up with a compelling solution. 😞

xolve:
There is no straight solution here. I would prefer to stick to `nvm` only. System-wide packages are not custom environments which user want to know per shell. Maybe raise an issue with jorgebucaran's plugin to add `current` to its method?

IlanCosman:
I apologize, I misinterpreted your previous message. Using `nvm current` seems like a reasonable solution. I'll commit some changes to your fork to make it conform to the newly created style guide and then hopefully merge. I'll also open an issue on jorgebucaran's plugin for `nvm current`.

IlanCosman:
Actually I think it would be simpler if I release 2.0 first and then merge this for a quick 2.1. 2.0 should come within a day or two (I hope).

xolve:
Seems reasonable :smile:

jorgebucaran:
Hey folks! I'm happy to add a `current` command to `nvm`, but I'd like to understand why and how it's useful first.

See: https://github.com/jorgebucaran/fish-nvm/issues/111#issuecomment-660831909

jorgebucaran:
> @xolve Approach by jorgebucaran' nvm plugin is to wrap around each nvm sub-command. IMHO it won't scale because you are always trying to catch-up with upstream. I wouldn't recommend it.

Whoops, this is wrong. [fish-nvm](https://github.com/jorgebucaran/fish-nvm) is a pure-fish, Node.js version manager implementation unrelated to `nvm-sh/nvm`. There's no "wrapping" and only fish and curl required. I would recommend reading the source or just the docs next time! 😉

IlanCosman:
Hi Jorge! The purpose of the `nvm current` command to is tell the user their current node version, replacing the system version with system. This tells the user if their sort of 'using' nvm or not. 

Often what a theme is trying to do is show you pertinent or unexpected information. The system node version is the normal expected thing, and if the user is using a different one we can likely infer that they are working on a node project for which they need a different version. 

I've never used node really so I'm not sure if a node display is particularly useful--it sounds like you would not! I'm simply following the lead of hundreds of other themes that have this feature, suggesting that some people do find it useful.

jorgebucaran:
Thank you; left you an answer here https://github.com/jorgebucaran/fish-nvm/issues/111#issuecomment-663889803. 👋 

> I'd just show the Node version, whatever it is. Managed or not makes no difference to me because switching to a particular version of Node with fish-nvm makes it immediately available and shared between all the current user's fish instances, and will be preserved across restarts of the shell as well. 
>
>I never used the original nvm, but I suspect it doesn't work like that. @ljharb or anyone else is welcome to correct me.
>
>---
>
> Referring "to system node" in my previous comment was a bit misleading. There's usually no system node unless you happen to have node installed already, e.g., you compiled/downloaded it yourself, and that's never been my use case, since I use fish-nvm to get node in the first place.
>
>Even if we add an `nvm current` command, it would be exactly the same as `node -v`. Let's leave the issue open for now.

xolve:
@jorgebucaran I read your code, like as an overview. Being new to `fish` looked like a customized wrapper, its my mistake.
Its good to see a native `fish` based `nvm` from scratch :clap: More options means more choice and perhaps a new functionality or perspective.

Personally I always blacklist distro's `node` packages because I have always been using `nvm`.

@IlanCosman I think we can switch to using `which node` and `node -v` to differentiate between system and `nvm` based `node` paths.

IlanCosman:
Release 2.0 is here! 😤 Thank you both so much for your patience @xolve and @jorgebucaran, I'm rather mortified by falling into the newbie trap of being too optimistic about timelines, I'll do better in the future. 😞 

Much appreciated for merging my project into awesome-fish @jorgebucaran 🙇 , I hadn't even realized until today ❤️ (I really like the new look of the readme btw) 

@xolve thanks again for your patience with your contribution , I'm excited to do a quick v2.1 release for this PR! 😃 To that end can you update your PR for v2.0, including the new [code contribution style guide](https://github.com/IlanCosman/tide/blob/master/CONTRIBUTING.md#code-conventions)? Thanks again 😄

IlanCosman:
> @IlanCosman I think we can switch to using `which node` and `node -v` to differentiate between system and `nvm` based `node` paths.

@xolve Could you explain this more? I don't quite follow you :(

jorgebucaran:
Congrats! 🎉🌊

xolve:
> > @IlanCosman I think we can switch to using `which node` and `node -v` to differentiate between system and `nvm` based `node` paths.
> 
> @xolve Could you explain this more? I don't quite follow you :(

We can use `node -v` to identify node version installed consistently. But this will not differentiate between system installed or `nvm` managed. To differentiate this we can use path returned by `which node`. Distro's `node` is usually installed in `/usr` while `nvm`'s is installed in home directory.

IlanCosman:
That sounds great 😃  Can you try it out? You'll need to update your repository to v2, I'm not sure how that works with the PR system. Thanks again for your patience with this 😄

IlanCosman:
Oh any by the way the tests do now apply in v2, and there's CI 💪😂

IlanCosman:
Hi there @xolve , renaming the default branch to main closed this PR automatically. If you're still interested in contributing, please open a new PR to main 😄

PR #23, opened 2020-09-06
Add nvm display to prompt.

xolve:
Display node version when using nvm.

#### Motivation and Context

People keep multiple node versions using nvm. This helps in better visibility of shell environment.

#### How Has This Been Tested

Tested with local `nvm`.

#### Checklist

Documentation has moved to wiki. Will update once PR is accepted.

~- [ ] I have updated the tests accordingly.
Waiting for first review. Adding test cases not easy without running in a container.~ *Unnecessary for now*

### DIFF
--- a/functions/_tide_item_nvm.fish
+++ b/functions/_tide_item_nvm.fish
@@ -1,23 +1,15 @@
-function _tide_node_version
-    function top_level_dir --argument-names dirPath
-	echo (string split --max 2 / $dirPath)[2]
-    end
-
-    set -l node_path_begin (top_level_dir (which node 2>/dev/null))
-    set -l home_path_begin (top_level_dir $HOME)
-
-    if test $node_path_begin = $home_path_begin;
-	echo (node --version)
-    else;
-	echo "system"
-    end
-end
-
 function _tide_item_nvm
-    set -l node_version (_tide_node_version)
+    if set -l whichNode (which node 2>/dev/null) # Ensure node is installed
+        set -l nodeVersion (node --version)
 
-    if test "$node_version" != "system" -a -n "$node_version"
-        set_color $tide_nvm_color
-        printf '%s ' "$tide_nvm_icon $node_version"
+        if test -n "$tide_nvm_default_node"
+            if test "$tide_nvm_default_node" != "$nodeVersion"
+                set_color $tide_nvm_color
+                printf '%s' {$tide_nvm_icon}' ' $nodeVersion
+            end
+        else if string match --quiet --regex "^$NVM_DIR.*" $whichNode # If node path begins with nvm directory
+            set_color $tide_nvm_color
+            printf '%s' $tide_nvm_icon' ' $nodeVersion
+        end
     end
 end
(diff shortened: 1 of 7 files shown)


IlanCosman:
It's great to see you again @xolve 😄 I felt badly about delaying your original PR for the major version bump. I'll review these changes and test them on MacOS. Good point about the documentation, I wish wikis were better integrated into their main repository.

xolve:
:smiley: Had been busy with changes at work and had to take a break from OSS explorations. Well lets finish it this time.

IlanCosman:
Thanks for the PR xolve. Your code was a really great starting point for the nvm item. Since I've rewritten most of the code, I'd like to discuss a few things about the original that you can improve on to create better fish scripts in the future :smile: In addition there were some [style guide][] elements that were lacking.

General things:

- There's no need to define `top_level_dir` inside `_tide_node_version`. Fish doesn't have private functions, so redefining `top_level_dir` every time is wasteful.
- Helper functions like `_tide_node_version` and `top_level_dir` should generally come after the main function.
- There isn't really a need for `_tide_node_version` to exist as a separate function. It is only run once and isn't long enough to be necessary to split off.

Details:

- There's no need to use `echo (node --version)`, that's the same as just using `node --version`.
- No need for a `;` after the `else`.
- When you are printing `tide_nvm_icon` and `node_version`, what happens if the user erases the icon variable? You'll be left with a space hanging before node version, which will look silly. Use `printf '%s' $tide_nvm_icon' ' $node_version`. If `tide_nvm_icon` is not defined, that entire argument will be erased, fixing the problem of the extra space. We don't need a space after `node_version` because `_tide_right_prompt` will take care of that automatically.

[Style guide][] things:

- The indentation is a bit wonky, for example with the `echo`s. You can use the builtin `fish_indent` command to fix that! There's also a [great VSCode extension][] for fish that will do this automatically for you and provide syntax highlighting.
- `echo` is considered to inferior to `printf`, which is why I discourage it in my project. [1] [2]
- Local variable names should be in `camelCase`.

Non-code stuff:

- I changed the default icon to the node logo: ⬢.
- The config files are organized alphabetically, so I moved the nvm variables to the correct place. I use the [sort lines][] VSCode extension for this.

Other changes:

- I added a new variable `tide_nvm_default_node`. This is useful for those who use Jorgebucaran's nvm.fish. nvm.fish doesn't have a conception of the "system" version, so `tide_nvm_default_node` can be used to prevent the nvm item from displaying when using a certain node version, set by the user.

---

I still need to test this item on MacOS.

As for automated tests, I'm planning to introduce command mocking soon that will make it much easier to write test for items like this, so we can ignore them for now.

Can you review and briefly tests my changes and see if they work for you?

I really appreciate your work, sorry for the wall of text :joy:

[1]: https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819
[2]: https://askubuntu.com/questions/467747/which-is-better-printf-or-echo
[great vscode extension]: https://marketplace.visualstudio.com/items?itemName=bmalehorn.vscode-fish
[sort lines]: https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines
[style guide]: https://github.com/IlanCosman/tide/blob/main/CONTRIBUTING.md#code-conventions

IlanCosman:
I've tested the item on MacOS and made a few changes. I can handle the documentation since I added a new variable and there are a few oddities to discuss. Thanks @xolve 😄

xolve:
Thanks for massive code clean up 😄 I do get the part where I missed to give option to set NVM_HOME.

PR #29, opened 2020-09-30
Explain variable modification for configuration

Epse:
Add a line to the documentation to clarify where one can add new items to the prompt manually.

#### Description

From the documentation it wasn't immediately clear to me where to find the configuration of the prompt and how to add new items. This PR adds a single line above the list of items to clarify where these can be added.

#### How Has This Been Tested

Documentation patch, no tests.

#### Checklist

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -14,7 +14,7 @@ Variables in these sections will be denoted as `foo_bar` but will actually be `t
 
 You can take a look at the default configurations in `~/.config/fish/tide_theme/configure/configs` for inspiration.
 
-Do not customize these variables directly in `fish_variables`, use `set --universal` or `set --global` in your `config.fish` file.
+You can modify variables using `set --universal` from the command line or `set --global` in your `config.fish` file.
 
 ## Prompt Variables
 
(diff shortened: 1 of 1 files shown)


IlanCosman:
Hi Epse, thanks for contributing 😄 

From the [docs](https://fishshell.com/docs/current/index.html?highlight=fish_variables#more-on-universal-variables):
> Universal variables are stored in the file `.config/fish/fish_variables`. Do not edit this file directly, as your edits may be overwritten. Edit the variables through fish scripts or by using fish interactively instead.

Instead of editing the `fish_variables` file, use `set --universal` from the commandline, or `set --global` on startup from your `config.fish`.

You are right that the documentation should do a better job of explaining how to modify the prompt, using the methods I explained above. Can you modify your addition to do this?

Epse:
Certainly, I just added a commit to do that!

Epse:
For the record, I have no clue what the prettier CI is complaining about. Prettier on my machine is perfectly happy.

Epse:
Alright managed to fix it

IlanCosman:
Thanks for contributing @Epse 😄

Epse:
You're welcome!

PR #39, opened 2020-10-23
Use TMPDIR

Shadow53:
This pull request replaces hard-coded references to `/tmp` with `$TMPDIR` to enable installation in contexts where `/tmp` is not
available, e.g. using Termux on Android.

#### Description

Hard-coded references to `/tmp` are replaced with `$TMPDIR`, and strings containing `/tmp` now use double quotes for proper
variable expansion. Before use, `TMPDIR`'s existence is tested and, if not found, set globally (script-local) to `/tmp`, for backwards-compatibility.

#### How Has This Been Tested

Because `$TMPDIR` is used during installation and certain paths are hardcoded, I tested the logic of my changes in the shell rather than testing the scripts themselves:

```fish
set -u TMPDIR
if ! set -q TMPDIR
    set -l TMPDIR /tmp
end
echo "$TMPDIR" # Nothing
if ! set -q TMPDIR
    set -g TMPDIR /tmp
end
echo "$TMPDIR" # "/tmp"
set -g TMPDIR /var/tmp
if ! set -q TMPDIR
    set -g TMPDIR /tmp
end
echo "$TMPDIR" # "/var/tmp"
```

#### Checklist

- [x] I have updated the documentation accordingly. (I do not think anything needs to be updated?)
- [x] I have updated the tests accordingly. (There do not look to be any installation tests, so nothing to update?)

### DIFF
--- a/functions/_tide_detect_os.fish
+++ b/functions/_tide_detect_os.fish
@@ -17,6 +17,7 @@ function _tide_detect_os
 end
 
 function _tide_detect_os_linux_cases -a file key
+    test ! -f $file && return 1
     set -l splitOsRelease (cat $file | string split '=')
     set -l value $splitOsRelease[(math (contains --index $key $splitOsRelease)+1)]
     set -l name (string trim --chars='"' $value | string lower)
--- a/functions/_tide_sub_test.fish
+++ b/functions/_tide_sub_test.fish
@@ -20,7 +20,7 @@ function _tide_sub_test
 
     if ! set -q TMPDIR
         set -g TMPDIR /tmp
-    endif
+    end
 
     set -lx TERM xterm # Necessary for testing purposes, ensures color codes are printed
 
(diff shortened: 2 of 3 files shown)


Shadow53:
Added a new commit because I somehow missed that I wrote `endif` instead of `end` when I made the changes. Must have been stuck in vimscript mode or something.

IlanCosman:
Hi @Shadow53 , thanks for the PR 😄 That's a good point, explicitly defining file names isn't a very good idea. I'm a bit nervous about setting `$TMPDIR` though, even if it's not already set. How about instead we use `mktemp`? Since it creates a new file each time, it should also be safer for `Ctrl+C` and other things like that. Are you willing to implement this? 

Also, if you could fix the merge conflicts that would be nice ❤️I think you mistakenly included some stuff from #40 in here.

Shadow53:
Yeah, I can do that. Since I accidentally included stuff from #40, I think the easiest option will be for me to start a new pull request based on the main branch. I'll tag this one in that one to mark them as related.

IlanCosman:
Sounds good 😄

PR #41, opened 2020-10-23
use mktemp instead of explicitly named files

Shadow53:
Version 2 of pull request #39. This time, it uses mktemp and the variable name `tempDir` (as found in existing code). These changes are in the `_tide_sub_test` and `_tide_actual_install` functions.

#### Description

The code was copied and pasted between both sections because I was not sure where it would be a good idea to put a function. It first checks if the version of mktemp is the GNU one, because it added the `--tmpdir` flag and deprecated `-t`, which is the preferred method on BSD/macOS systems, then runs the correct invocation to get the path.

If creating the temporary directory fails, both functions will exit.

`tempDir` is created as a global variable in both instances and is unexported with `set -u` in `_tide_actual_install` to attempt to prevent issues for the user.

#### How Has This Been Tested

I tested the individual pieces of the code manually in my shell. Temporary directory creation works using the GNU version; BSD was untested but assumed to work since it was written based on the man page(s).

I ran the tests using the modified version of `_tide_sub_test` and there were no visible issues.

#### Checklist

I do not think there is documentation to update.

### DIFF
--- a/functions/_tide_sub_test.fish
+++ b/functions/_tide_sub_test.fish
@@ -22,9 +22,9 @@ function _tide_sub_test
 
     set -l testsDir "$_tide_dir/tests"
 
-    set -l pending '/tmp/tide_test'
-    set -l failed '/tmp/tide_test_failed'
-    set -l passed '/tmp/tide_test_passed'
+    set -l pending (mktemp)
+    set -l failed (mktemp)
+    set -l passed (mktemp)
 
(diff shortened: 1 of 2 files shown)


IlanCosman:
I simplified your code a bit. There's no need for it to be called `tide.blah`. Thus we can simply used `mktemp` and `mktemp -d`.

PR #45, opened 2020-11-11
Move fish_right_prompt to fish_right_prompt.fish

jorgebucaran:
If we keep the `fish_right_prompt` function in `fish_prompt`, it stays defined in the session even if one uninstalls Tide.


IlanCosman (approved):

IlanCosman:
Good catch! I remember noticing this and then not doing anything about it 😅

Issue #62, opened 2020-12-30
Tide tests assume that tide is installed in $__fish_config_dir


vamega:
#### Describe the bug

Right now some code assumes that tide is installed under `$__fish_config_dir`.
* [tests/context.fish](https://github.com/IlanCosman/tide/blob/d1b2806977f14daa6987dd05c776c9a5159cc12e/functions/tide/tests/context.fish#L3)
* [_tide_init.fish](https://github.com/IlanCosman/tide/blob/1bcfe317e4c4feb408372309143a70f6b76219aa/conf.d/_tide_init.fish#L8-L14)

If installed via fisher it is possible that tide is actually under a different directory root, through the use of `$fisher_path`. See [this fisher issue](https://github.com/jorgebucaran/fisher/issues/640) for more details. 

#### Steps to reproduce

Don't really know the best way to reproduce this, since I'm not familiar with how to run the tests. Should the tests be run without `Tide` installed, or is it okay to run them in a shell where `Tide` is already running?

1. Set $fisher_path to ~/.config/fish/fisher
2. Place the snipped from [here](https://github.com/jorgebucaran/fisher/issues/640#issue-771658021) in `~/config/.fish/conf.d/fisher.fish`
2. Install fisher
3. Install tide
4. See error about sourcing `_tide_sub_configure.fish`

I don't the setup for the tests (Do you run the tests after you've installed tide, or separately?), but I think the test above should fail.

#### Expected behavior

The tests should not assume an installation path, but instead discover it's installation path, and identify paths under that root.

#### Environment (please complete the following information)

- Operating System: Arch Linux
- Output of `tide bug-report`: 

```console
fish, version 3.1.2
tide, version 3.0.0
xterm-256color
```

#### Additional context

I had a fork that was addressing some of these issues using a universal variable, but it seems like you've removed almost all instances of this assumption. This test case is the only one left, but it seems like you've mostly eliminated the issue with a different approach, so if you could fix this one last test I think you'll have completely removed the issue.

jorgebucaran:
Just a thought. Fisher could pass the current path to `*_install` events so that we don't need to use `$__fish_config_dir` or `$fisher_path` here:

https://github.com/IlanCosman/tide/blob/1bcfe317e4c4feb408372309143a70f6b76219aa/conf.d/_tide_init.fish#L1-L14

Or maybe use `status filename` to get the current file name and extract the directory from there.

vamega:
[...]

PR #65, opened 2021-01-01
Allow for installation into non-standard locations

vamega:
This is useful when installed via fisher, if the user has set up
$fisher_path to be something other than ~/.config/fish

Closes: #62

#### Description
Tide now works when installed into places other the default `$fisher_path` under `$__fish_config_dir`.

Has a custom implementation of that I think is equivalent of `(dirname (dirname (<path>))`, but I haven't run extremely through tests on it.

#### How Has This Been Tested

I ran `tide test --all` on my machine. Some TAP tests failed, but inspecting those changes I think they're due to my configuration, and I expect them to pass in CI.

#### Checklist


conf.d/_tide_init.fish
```diff
@@ -1,17 +1,28 @@
+# Gets the parent of the parent of a given path
+# Returns the current directory if the given path does not have two parents
+function _tide_dirname_2 -a somepath
```
jorgebucaran, line 3: Do we really need to call this `_tide_dirname_2`? It seems to me that just `_tide_dirname` would suffice.

IlanCosman:
Thanks for this PR Varun 😄 I managed to find some simpler regex to do the trick, but I really appreciate you going out of your way to fulfill my somewhat silly request 😄

jorgebucaran:
@IlanCosman In case you're interested, here's a shorter version:

```fish
string join / (status filename | string split /)[1..-3]
```

IlanCosman:
Yah I thought of that. While it works for this exact case, it's not an exact dirname equivalent.

```fish
ilan@arch ~> string join / (string split / /home/ilan/fish)[1..-3]
/home
ilan@arch ~> string join / (string split / /home/ilan/fish/)[1..-3]
/home/ilan
ilan@arch ~> dirname /home/ilan/fish
/home/ilan
ilan@arch ~> dirname /home/ilan/fish/
/home/ilan
```

vamega:
@IlanCosman it seems your simpler regex also doesn't do what `dirname` does.

Does not seem to work when the path has a trailing slash. This shouldn't be the case in the usage here, since we're using it on a file, and it does not result in a path with a trailing slash. The implementation that was merged in is actually worse than the one that @jorgebucaran posted in this sense..

```
> echo /home/vmadiath/abc/ | string replace --regex '/[^/]+$' ''
/home/vmadiath/abc/

> dirname /home/vmadiath/abc/
/home/vmadiath
```

It also doesn't seem to share the behaviour where the path has extraneous path separators. This again shouldn't be a problem, since I don't think that `status filename` would return a path in this form even though it would be valid.

```
> echo /home/vmadiath/abc///123.txt | string replace --regex '/[^/]+$' '' | string replace --regex '/[^/]+$' ''
/home/vmadiath/abc//
> dirname (dirname /home/vmadiath/abc///123.txt)
/home/vmadiath
```

I'm not saying that we need the complexity of an exact `dirname` equivalent. Turns out getting the edge cases right isn't straightforward.

IlanCosman:
Nice catch. Yeah this is a tough nut to crack, but as you said, not really an important one 😄

IlanCosman:
Alright, I _think_ I've got it 😂 `string replace --regex '(/{2,})?[^/]+/*$' ''`

IlanCosman:
Alright alright _THIS_ time I know I've got it. `string replace --regex '(?!^/)(/{2,})?[^/]+/*$' ''`

vamega:
I think my regex/function actually does a pretty good job.
I'm not sure as to the feasibility of accurately doing this with just a regex.
In some cases you'll need to replace the input with a `.` For example

```
> dirname home
.
```

If you really want to beat this horse, I could write up a [hypothesis](https://hypothesis.readthedocs.io/en/latest/) test to test equality.

IlanCosman:
I think I'm actually damn close with that last regex 😂 But no, I'll stop, the path to regex hell is paved with good intentions. We'll leave it how it is until fish 3.2 comes out and we can discard it in favor of `status dirname`.

vamega:
You'll still need a `dirname` on top of that.
We were applying `dirname` once to rid ourselves of the filename, and once
to rid ourselves of the parent.
Is there a scenario you can think of where my regex doesn't work?

Obligatory reference: https://xkcd.com/1171/

IlanCosman:
Sorry for the late reply @vamega. You're totally right, I misunderstood what `status dirname` was supposed to do.

The embarrassing truth is that I didn't look at your function very carefully. I thought to myself "there's no need for a whole new function here." It's simple and only used in one place. Also, your regex seemed overly complex for the job, though as I quickly learned it was only somewhat so 😂 I still agree with both points, but, in arrogance perhaps, I dismissed your work too quickly and for that I apologize.

I actually do think that last regex works. Replacing the current directory with `.` isn't important, because `status current-filename` won't give things with dots in them.

Issue #78, opened 2021-01-18
It barfs with miniconda


wirtsi:
#### Describe the bug

tide doesn't like the way miniconda is hooking itself

#### Steps to reproduce
Add these lines to config.fish
```
>>> conda initialize >>>
!! Contents within this block are managed by 'conda init' !!
eval /Users/wirtsi/miniforge3/bin/conda "shell.fish" "hook" | source
<<< conda initialize <<<
```
And you get these errors

```
~/.config/fish/config.fish (line 46): Expected a command, but instead found a redirection
>>> conda initialize >>>
^
from sourcing file ~/.config/fish/config.fish
	called during startup
source: Fehler beim Lesen der Datei '/Users/wirtsi/.config/fish/config.fish'
Welcome to fish, the friendly interactive shell
~/.config/fish/config.fish (line 46): Expected a command, but instead found a redirection
>>> conda initialize >>>
^
from sourcing file ~/.config/fish/config.fish
	called during startup
~/.config/fish/config.fish (line 46): Expected a command, but instead found a redirection
>>> conda initialize >>>
^
from sourcing file ~/.config/fish/config.fish
	called during startup
source: Error while reading file '/Users/wirtsi/.config/fish/config.fish'
source: Error while reading file '/Users/wirtsi/.config/fish/config.fish'
~/.config/fish/config.fish (line 46): Expected a command, but instead found a redirection
>>> conda initialize >>>
^
from sourcing file ~/.config/fish/config.fish
	called during startup
source: Error while reading file '/Users/wirtsi/.config/fish/config.fish'
 ~ ~/.config/fish/config.fish (line 46): Expected a command, but instead found a redirection                                                                                                                                                            ✔ 
>>> conda initialize >>>
^
from sourcing file ~/.config/fish/config.fish
	called during startup
source: Error while reading file '/Users/wirtsi/.config/fish/config.fish'
```

The conda init script outputs this and pipes it to fish, I guess there it clashes with the right prompt setup

```
set -gx CONDA_EXE "/Users/wirtsi/miniforge3/bin/conda"
set _CONDA_ROOT "/Users/wirtsi/miniforge3"
set _CONDA_EXE "/Users/wirtsi/miniforge3/bin/conda"
set -gx CONDA_PYTHON_EXE "/Users/wirtsi/miniforge3/bin/python"
# Copyright (C) 2012 Anaconda, Inc
# SPDX-License-Identifier: BSD-3-Clause
#
# INSTALL
#
[...]

PR #84, opened 2021-02-02
Fix errors displayed when using tide prompt with conda

omad:
This fixes a spam of warning messages which print out after every prompt, when using tide together with [conda](https://github.com/conda/conda)

This has been working on Linux and MacOS for me for the last month.

**Example warning messages printed out after every prompt:**
```

~/.config/fish/functions/fish_prompt.fish (line 30): 
    ^
in function '_tide_refresh_prompt' with arguments 'VARIABLE SET _tide_right_prompt_display_48151'
	called on line 1 of file ~/.config/fish/functions/_tide_right_prompt.fish
in event handler: handler for variable “_tide_right_prompt_display_48151”
	called on line 24 of file ~/.config/fish/functions/_tide_right_prompt.fish

(Type 'help commandline' for related documentation)
```

#### Checklist

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -27,7 +27,7 @@ function fish_prompt
 end
 
 function _tide_refresh_prompt --on-variable _tide_left_prompt_display_$fish_pid --on-variable _tide_right_prompt_display_$fish_pid
-    commandline --function force-repaint
+    commandline --function force-repaint 2>/dev/null
 end
 
 # Double underscores to avoid erasing this function on uninstall


IlanCosman:
Have you taken a look at https://github.com/IlanCosman/tide/issues/78#issuecomment-762374266? Does that fix this issue for you?

omad:
I hadn't seen that, thanks, it does fix it too.

Is this something that **conda** is doing incorrectly, and should a fix be pushed there?

IlanCosman:
This is a [Fish issue](https://github.com/fish-shell/fish-shell/issues/5394). Basically, Fish reads `config.fish` even when the shell isn't being run interactively. This causes your conda init statement to run even in the background job. Conda init does this thing where they [add an indicator](https://github.com/conda/conda/blob/aed799b56d9ee16a3653928527e2af6e41394e85/conda/shell/etc/fish/conf.d/conda.fish#L9-L54) to the user's prompt, which causes `fish_prompt.fish` to be loaded, which causes the issues you're seeing because Tide isn't meant to be used non-interactively. 

For now, the best solution is simply to add the `status is-interactive` to your `config.fish`. Perhaps I should add something to the readme to this effect.

I really appreciate you taking the time to put a PR together and share your solution with others 😄

PR #92, opened 2021-02-13
Add ruby item

pusewicz:
#### Description

This change adds the Ruby version item to the prompt.

#### Motivation and Context

As a Ruby developer, it's great to see what version of Ruby I'm currently using.

image

#### How Has This Been Tested

I have tested manually on my machine and added tests.

#### Checklist

Question: How would I go about disabling the Ruby icon when my prompt has icons disabled? I'd like to add that. I've set the Nerd Font glyph correctly, but I do not currently use any kind of Nerd Font in my terminal.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -75,6 +75,7 @@ set --universal tide_left_prompt_items context $tide_left_prompt_items
 | [os](#os)                     | current operating system          |
 | [prompt_char](#prompt_char)   | prompt symbol; turns red on error |
 | [pwd](#pwd)                   | current directory                 |
+| [ruby](#ruby)                 | ruby version number               |
 | [rust](#rust)                 | rust version number               |
 | [status](#status)             | exit code of the last command     |
 | [time](#time)                 | current time                      |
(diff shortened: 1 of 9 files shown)


IlanCosman:
Wow! Thanks for the PR 😄 Item, tests, and documentation. Solid! 👍 

To disable the Ruby icon, you'd set it to empty with `set tide_ruby_icon`. Note that it has to be empty, not empty string. Tide uses Fish's cartesian expansion system to remove the whole term when one element evaluates to literal nothing. You did this with `printf '%s' $tide_ruby_icon' ' (some other stuff...)`. This should probably be in the docs 😅

Honestly though, you should probably just have a nerd font enabled in your terminal. Tide still has many icons by default, and turning them all off would be annoying. If you're worried about not having the font you like, Nerd fonts has 51 patched fonts, so your favorite is probably on the list 😄 Eventually, an ascii mode like what powerlevel10k has would be nice, to solve your issue.

I'm currently on a roadtrip with family, so it might take me a little bit to get this merged. Thanks a ton for this contribution 😄

pusewicz:
Haha, unfortunately Envy Code R is _not_ 😭. I wonder how difficult would it be to patch it myself.

pusewicz:
Actually, just noticed, that my Ruby version is not correct:

```sh
~/Work/project1 pu-node-version *26  3.0.0p0 ·········································· 19:33:34
❯ _tide_item_ruby
 2.7.2p137⏎

~/Work/project1 pu-node-version *26  3.0.0p0 ·········································· 19:33:36
❯ which ruby
/Users/piotr/.rubies/ruby-2.7.2/bin/ruby
```

It shows the version of Ruby that I have set as a system-wide.

PR #93, opened 2021-02-13
[tide]Different connection color if root or ssh

claneys:
#### Description

Change the color and style of the prompt connection characters depending on the context.

#### Motivation and Context

Got this previously on my bash prompt and found that useful as the user information is at the right and might be not catched at first glance. Setting the color obviously showing you that you are using the root account or that you are on a remote connection.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_prompt.fish
+++ b/functions/_tide_prompt.fish
@@ -4,7 +4,13 @@ function _tide_prompt
 
     test "$tide_print_newline_before_prompt" = 'true' && printf '%b' '\n'
 
-    set_color $tide_prompt_connection_color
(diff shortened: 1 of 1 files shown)


IlanCosman:
While the context item has bright colors, I can definitely see how it might be overlooked all the way on the right side. However, if this is a problem for you, you can move it to the left side pretty easily (you can read about that in the configuration guide).

Because I don't think this is a problem for the majority of people, and because there exists a reasonable solution, I'm going to close this PR. However, I really appreciate you taking the time to contribute 😄

PR #98, opened 2021-02-16
Add support for poetry and conda virtualenvs

### LINKED ISSUE #97: virtualenv is displayed incorrectly for poetry and conda
When using [poetry](https://python-poetry.org/) to manage virtualenvs, this is what tide displays

image

The `virtualenvs` should be the name of my virtualenv.

```bash
echo $VIRTUAL_ENV
/home/user/.cache/pypoetry/virtualenvs/myproject-j_EVlIcP-py3.8
```

So the venv should be displayed as `myproject-j_EVlIcP-py3.8` or `myproject`. Something similar could be implemented for [conda](https://docs.conda.io/en/latest/) users. Here, `CONDA_DEFAULT_ENV` should be used instead of `VIRTUAL_ENV`.

I'll cook up a pull request for [`_tide_item_virtual_env.fish`](https://github.com/IlanCosman/tide/blob/main/functions/_tide_item_virtual_env.fish) in a bit!

chvolkmann:
### Description

Adds support for conda envs and proper displaying of poetry virtualenvs.

### Describe your changes

Updated `_tide_item_virtual_env`.

If `VIRTUAL_ENV` is not set, it will be checked if `CONDA_DEFAULT_ENV` is set

If poetry is detected, we pick the last part of the virtualenv part as the name.
```
/home/user/.cache/pypoetry/virtualenvs/some-project-j_EVlIcP-py3.8
```

Within the last part, we split by `-` and throw away the last two elements.

### Why is this change required? What problem does it solve? 
Closes #97

### How Has This Been Tested
See `tests/virtual_env.fish`. For me, the tests I changed ran fine, but the rust test failed for some reason.

```
$ python3 ./littlecheck/littlecheck/littlecheck.py ./tests/**
Failure in ./tests/rust.fish:

  The CHECK on line 16 wants:
    \xe7a8 1.30.0-beta

  but there was no remaining output to match.

  Context:
     <= nothing to match CHECK '\xe7a8 1.30.0-beta' on line 16
     <= nothing to match CHECK '\xe7a8 1.30.0' on line 23

  when running command:
    fish ./tests/rust.fish
```

#### Checklist

### DIFF
--- a/functions/_tide_item_virtual_env.fish
+++ b/functions/_tide_item_virtual_env.fish
@@ -2,10 +2,24 @@ function _tide_item_virtual_env
     if set -l splitVirtualEnv (string split '/' "$VIRTUAL_ENV")
         set_color $tide_virtual_env_color
 
-        if test "$tide_virtual_env_display_mode" = 'venvName'
-            printf '%s' $tide_virtual_env_icon' ' $splitVirtualEnv[-1]
-        else # Default to projectName
-            printf '%s' $tide_virtual_env_icon' ' $splitVirtualEnv[-2]
+        # Detect if the virtualenv was spawned by poetry
+        # then _OLD_VIRTUAL_PATH is set and VIRTUAL_ENV has the form
+        # /home/<user>/.cache/pypoetry/virtualenvs/myproject-j_EvlIcP-py3.8
+        if test -n "$_OLD_VIRTUAL_PATH"; and test $splitVirtualEnv[-3] = 'pypoetry'
+            # This is needed when the project name has a dash in it, e.g. fish-shell
+            # Split by - and discard the last two parts
+            set -l splitPoetryVirtualEnv (string split '-' $splitVirtualEnv[-1])
+            set -l poetryVirtualEnvName (string join '-' $splitPoetryVirtualEnv[1..-3])
+            printf '%s' $tide_virtual_env_icon' ' $poetryVirtualEnvName
+        else
+            if test "$tide_virtual_env_display_mode" = 'venvName'
+                printf '%s' $tide_virtual_env_icon' ' $splitVirtualEnv[-1]
+            else # Default to projectName
+                printf '%s' $tide_virtual_env_icon' ' $splitVirtualEnv[-2]
+            end
         end
+    else if test -n "$CONDA_DEFAULT_ENV"
+        # conda uses CONDA_DEFAULT_ENV instead of VIRTUAL_ENV
+        printf '%s' $tide_virtual_env_icon' ' $CONDA_DEFAULT_ENV
     end
 end
--- a/tests/virtual_env.fish
+++ b/tests/virtual_env.fish
@@ -20,3 +20,14 @@ _virtual_env # CHECK:  .venv
 set -lx tide_virtual_env_icon
 set -lx tide_virtual_env_display_mode projectName
 _virtual_env # CHECK: python_project
+
+# Support for poetry virtualenvs
+set -lx VIRTUAL_ENV /home/user/.cache/pypoetry/virtualenvs/project-j_EVlIcP-py3.8
+_virtual_env # CHECK: project
+set -lx VIRTUAL_ENV /home/user/.cache/pypoetry/virtualenvs/some-project-j_EVlIcP-py3.8
+_virtual_env # CHECK: some-project
+
+# Support for conda virtualenvs
+set -lx VIRTUAL_ENV
+set -lx CONDA_DEFAULT_ENV base
+_virtual_env # CHECK: base
--- a/tests/virtual_env.fish
+++ b/tests/virtual_env.fish
@@ -22,6 +22,7 @@ set -lx tide_virtual_env_display_mode projectName
 _virtual_env # CHECK: python_project
 
 # Support for poetry virtualenvs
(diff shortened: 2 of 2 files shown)


IlanCosman:
Thanks for the contribution! It's always nice when someone sees a spot for improvement and takes the opportunity 👍 

@chvolkmann @pusewicz
This and #92 require some further consideration. Powerlevel10k has seperate items for each tool, `asdf`, `venv`, `anaconda`, `pyenv`, etc. Spacefish operates this way for Python, but then for Ruby it has one item that interacts with all the relevant tools, `rbenv`, `rvm`, etc. The second seems a bit more elegant to me, but that's not a very good reason for doing it 😂 Could I ask your thoughts on the matter?

chvolkmann:
I don't think there is a real use case for showing multiple active virtualenvs in Python. But the way spacefish handles it is nice I think, differentiating between conda and venv.

Haven't used Ruby in a while, so it's hard to comment on that. :)

pusewicz:
After opening my PR I'm thinking that I'm leaning more towards per-tool item. So instead of a `ruby` item, I think I'll spend some time changing my PR and moving it to a `chruby` tool. That way the user is very explicit about the tool they are using and `tide` does not have to run a collection of tools to figure out how one's Ruby version is controlled. This saves battery, time, heat, and noise of the fans.

IlanCosman:
@pusewicz Well that leads into a question of defaults. You seem to be in favor of/assuming that we'd have these items off by default, and then the user will add them when necessary. I think there is a discoverability issue there. 

For example, right now, the `rust` item is off by default (simply because it is that way in P10K). How do we expect Rust programmers to find that item and enable it? Unlike P10K, we don't have a configuration file with commented out items that will catch their eye. Sure, some users will go look at the documentation or see the `_tide_item_rust.fish` file. But I imagine a large majority will simply use the configuration wizard and be done.

Thus, Tide needs to anticipate the user's needs. It's unergonomic for the user to have to think "It would be really nice if there was a rust item," and then go look at Tide's documentation.

So, in my opinion:

- Dumb solution - Have all such items on by default. `command -q` takes ~100 microseconds on my machine. If we have 10 items like this, that's an extra millisecond. Add in some extra overhead for fetching the items, etc. and we are looking at an 5-10% slowdown by my estimate. Not great, but not terrible.

- Smart solution - Have all such items off by default. Then, during `tide configure`, we detect that the user has a certain tool installed and enable the relevant item accordingly. Also do the same on interactive startup, except without the automatic-enabling because the user might have a custom configuration.

Thoughts? Sorry if this is too much text 😅

chvolkmann:
> Thus, Tide needs to anticipate the user's needs. It's unergonomic for the user to have to think "It would be really nice if there was a rust item," and then go look at Tide's documentation.

I totally agree, I've come to face that this is somewhat unpractical myself.


> Dumb solution - Have all such items on by default. command -q takes ~100 microseconds on my machine. If we have 10 items like this, that's an extra millisecond. Add in some extra overhead for fetching the items, etc. and we are looking at an 5-10% slowdown by my estimate. Not great, but not terrible.

That brings forward the question - do we want to have _everything_ on by default? Are some plugins mutually exclusive to one another? Also I'm not sure if that's the best idea considering that there can be a lot of items for a small amount of screen space.

> Smart solution - Have all such items off by default. Then, during tide configure, we detect that the user has a certain tool installed and enable the relevant item accordingly. Also do the same on interactive startup, except without the automatic-enabling because the user might have a custom configuration.

That sounds more reasonable.

> How do we expect Rust programmers to find that item and enable it? Unlike P10K, we don't have a configuration file with commented out items that will catch their eye. Sure, some users will go look at the documentation or see the _tide_item_rust.fish file. But I imagine a large majority will simply use the configuration wizard and be done.

This might be a bit involved, but I'd suggest **extending the configuration wizard for more fine-grained control**. Maybe a mode in which all possible features, i.e. all config variables, are listed with their current value. The user should then be able to enable/disable/set a feature easily, either by a direct command or by pressing button X in the wizard. 

```
$ tide wizard-ex
-- Python --
[1] Show Python version - yes
[2] Show virtualenv name - yes
[3] Show conda env name - no

-- NodeJS --
[4] Show nvm version - yes

....
```

IlanCosman:
@chvolkmann I don't think an advanced mode for the configuration wizard would be particularly useful. If you're customizing beyond the wizard, it is expected that you will look at the documentation. Tool items are a special case where Tide can anticipate the user's needs.

@chvolkmann and @pusewicz: I've coded up the auto-enabling feature (really it's auto-disabling) in the `v4.1` branch. I feel bad asking, but would you guys be willing to open new PRs targeting that branch? There are a few relevant differences:

1. Different (improved) test file naming scheme
2. All tool-items should be "on" by default, and are then removed if the machine doesn't have the required tool installed. So you should add your items to `tide_right_prompt_items`. 

@chvolkmann: Given the new structure, could you split up this PR into 2 PRs, each for a new item? 1 for poetry and 1 for conda.

@chvolkmann and @pusewicz: Sorry to put you guys in this annoying position 😞 Tide should have already had this structure, so that your PRs would be smooth sailing. If you guys don't want to update your PRs, I'm still super grateful that you took the time to contribute, and will of course attribute you in the release notes ❤️

PR #99, opened 2021-02-16
Add option for shorter hostname

chvolkmann:
### Description

Host names can be rather long for some machines. For instance, if your machine has a [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) set, the `$hostname` variable evaluates to `mymachine.mydomain.tld`, which can take up a lot of space in the terminal.

This pull request proposes a configuration option `short_hostname`, which strips away the domain end of the hostname.  
So for an FQDN of `mymachine.mydomain.tld`, we only display `mymachine`.

### How Has This Been Tested

### Checklist

- [ ] IPs are handled correctly (is this needed?)

### DIFF
--- a/functions/_tide_item_context.fish
+++ b/functions/_tide_item_context.fish
@@ -1,12 +1,20 @@
 function _tide_item_context
+    set -l displayedHost $hostname
+
+    if test "$tide_short_hostname" = 'true'
+        # note: might break for IPs
+        # TODO: regex for IPv4/IPv6
+        set displayedHost (string join '.' (string split '.' $displayedHost)[1..-3])
+    end
+
     if set -q SSH_TTY
         set_color $tide_context_ssh_color
-        printf '%s' $USER'@'$hostname
+        printf '%s' $USER'@'$displayedHost
     else if test $USER = 'root'
         set_color $tide_context_root_color
-        printf '%s' $USER'@'$hostname
+        printf '%s' $USER'@'$displayedHost
     else if test "$tide_context_always_display" = 'true'
         set_color $tide_context_default_color
-        printf '%s' $USER'@'$hostname
+        printf '%s' $USER'@'$displayedHost
     end
 end


IlanCosman:
I'm really sorry I forgot to reply to this 😅 I was a little hesitant because I couldn't find any other prompt that did this, but it seems that starship does offer something similar.

functions/_tide_item_context.fish
```diff
@@ -1,12 +1,20 @@
 function _tide_item_context
+    set -l displayedHost $hostname
+
+    if test "$tide_short_hostname" = 'true'
+        # note: might break for IPs
+        # TODO: regex for IPv4/IPv6
+        set displayedHost (string join '.' (string split '.' $displayedHost)[1..-3])
```
IlanCosman, line 7: We can make this `set displayedHost (string split '.' $displayedHost)[1]`

IlanCosman (changes requested):

functions/_tide_item_context.fish
IlanCosman: Also no need for the comments

functions/_tide_item_context.fish
```diff
@@ -1,12 +1,20 @@
 function _tide_item_context
+    set -l displayedHost $hostname
+
+    if test "$tide_short_hostname" = 'true'
```
IlanCosman, line 4: This variable should be `tide_context_short_hostname`

IlanCosman (changes requested):
The new option variable should be added to the configs, turned off by default of course.

functions/_tide_item_context.fish
IlanCosman: Variable should also be added to the wiki documentation, though of course that can't be in the PR 😞

IlanCosman:
I'm going to close this for now. I'd welcome a new PR in the future if you're still interested @chvolkmann 😄

PR #106, opened 2021-02-26
Refresh only if session is interactive

chvolkmann:
#### Description

After installing [miniconda](https://docs.conda.io/en/latest/miniconda.html) on my machine and initializing the shell using `conda init fish`, tide throws errors with every command.

```
╭─ ~/devel                                                                                                                                                            ✔  13:07:21 
╰─commandline: Can not set commandline in non-interactive mode                                                                                                                (py38) 

~/.config/fish/functions/fish_prompt.fish (line 33): 
    ^
in function '_tide_refresh_prompt' with arguments 'VARIABLE SET _tide_right_prompt_display_4029952'
        called on line 1 of file ~/.config/fish/functions/_tide_right_prompt.fish
in event handler: handler for variable “_tide_right_prompt_display_4029952”
        called on line 27 of file ~/.config/fish/functions/_tide_right_prompt.fish

(Type 'help commandline' for related documentation)

~/.config/fish/functions/fish_prompt.fish (line 33): 
    ^
in function '_tide_refresh_prompt' with arguments 'VARIABLE ERASE _tide_right_prompt_display_4029952'
        called on line 1 of file ~/.config/fish/functions/fish_prompt.fish
in event handler: handler for variable “_tide_right_prompt_display_4029952”
        called on line 39 of file ~/.config/fish/functions/fish_prompt.fish

(Type 'help commandline' for related documentation)
```

To fix this, I simply added a check to see whether we're currently interactive or not using [`status is-interactive`](https://fishshell.com/docs/current/cmds/status.html).

Related issue: [fish-shell#5394](https://github.com/fish-shell/fish-shell/issues/5394)

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -5,13 +5,16 @@ set -g _tide_left_prompt_display_var _tide_left_prompt_display_$fish_pid
 
 set -gx _tide_fish_pid $fish_pid
 
-function fish_prompt
-    set -lx _tide_last_pipestatus $pipestatus
 
-    if not set -e _tide_repainting
-        set -lx _tide_jobs_number (jobs --pid | count)
 
-        fish --command "
+if status is-interactive
+    function fish_prompt
+        set -lx _tide_last_pipestatus $pipestatus
+
+        if not set -e _tide_repainting
+            set -lx _tide_jobs_number (jobs --pid | count)
+
+            fish --command "
         set CMD_DURATION $CMD_DURATION
         set COLUMNS $COLUMNS
         set fish_bind_mode $fish_bind_mode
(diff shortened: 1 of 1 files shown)


IlanCosman:
I was just thinking about this problem. 👍

I'm a bit torn. On the one hand, erroring like this is uncool. On the other, it's indicative of an inefficiency in the user's configuration. Initializing conda presumably takes some time, which gets directly increases Tide latency.

It's probably still worth solving, at least until Fish offers a way of not loading the user's configuration.

Thanks for making a PR to solve this 😄 I think we can find a better solution though. This involves running `status is-interactive` on every prompt refresh, which is rather wasteful. Could you instead wrap the whole `fish_prompt.fish` file in an `if status is-interactive`, with an `else` that defines an empty `fish_prompt`?

IlanCosman:
Also, thanks for the thoughtfulness of targeting the `dev` branch ❤️ Unfortunately, in this case it didn't turn out well, because I intended to scrap that branch's latest commit which caused the tests to fail 😂

chvolkmann:
> Also, thanks for the thoughtfulness of targeting the dev branch ❤️ Unfortunately, in this case it didn't turn out well, because I intended to scrap that branch's latest commit which caused the tests to fail 😂

Haha well, it was out of good intent at least 😄   
Should I retarget another branch to merge into?

> Could you instead wrap the whole fish_prompt.fish file in an if status is-interactive, with an else that defines an empty fish_prompt?

Done

IlanCosman:
Sorry to micromanage 😅 Wrapping the whole thing like this would be better:

```fish
if status is-interactive
  _tide_detect_os
  etc...
else
    # Do nothing while not interactive
    function fish_prompt
    end
end
```

If we're unfortunately allowing that file to be sourced in the background process, we should minimize it's impact when doing so, and `_tide_detect_os` is a bit slow.

> Should I retarget another branch to merge into?

Yah, main would be good

PR #107, opened 2021-02-26
Refresh only if session is interactive 

chvolkmann:
Continuation of #106 to merge into `main` instead of `dev`


chvolkmann:
Switched around the order of if and else. Does this work?

IlanCosman:
Looks great 😄 Thanks Christian!

PR #115, opened 2021-03-05
Get branch name on older version of git (compatible with Debian Buster)

AdrianTM:


IlanCosman (changes requested):
the `if test -n "$branch"` on line 13 must be changed to `if test "$branch" != HEAD` because `git-rev-parse` doesn't return empty if not on a branch.

IlanCosman:
Apparently that's not the reason the tests failed. 😂 Was still necessary though. I'll investigate the tests. Thanks for the PR 😄

IlanCosman:
Ah, you wrote `git rev-parse --abbrev-ref`, instead of `git rev-parse --abbrev-ref HEAD`. Tests doing their job for one 😂

AdrianTM:
Grrr... Thanks for the patience!

IlanCosman:
Tests doing their job again 😄 This fails in a perfectly new git repository.

IlanCosman:
Huh, this is actually not so simple. https://www.techiedelight.com/determine-current-branch-name-git/

4 ways of doing it all have drawbacks.

1. `git branch --show-current` fails if the git version is too old
2. `git rev-parse --abbrev-ref HEAD` fails in a new git repo
3. `git symbolic-ref --short HEAD` fails if we aren't on a branch, so we can only use it if we put the next `git rev-parse` in front of it.
4. `git name-rev --name-only HEAD` is just too annoying to work with, you'll see what I mean if you try it out

IlanCosman:
Ok, yeah, so `git symbolic-ref --short HEAD` is out of the running since putting the next `git rev-parse` in front doesn't work, since that `git rev-parse` might fail even if we're in a git repo.

So it's just a decision between `git branch --show-current` vs `git rev-parse --abbrev-ref HEAD`.

So which is more important, working in a new git repo, or working on Debian (every other distro has a git version that works).

AdrianTM:
Can we handle the edge case of a new repo in a different way god a test if "git branch" returns an empty string for example? and then use the git rev-parse --abbrev-ref HEAD

AdrianTM:
Something like this? 

>git branch | string length -q && git rev-parse --abbrev-ref HEAD

or maybe 

>git branch | string length -q &&  git rev-parse --abbrev-ref HEAD; true 

(not sure how pipe stuff propagate in fish, I'm new at this)

AdrianTM:
Still errors out, I give up for now.

IlanCosman:
It would be `set branch (git branch --show-current 2>/dev/null || git rev-parse --abbrev-ref HEAD 2>/dev/null)` I believe. That would slow down every prompt run though, because I use `git branch` failure to tell whether we are in a git repo at all.

Thanks for trying, but I think I won't change this. The fact of the matter is that git 2.22 (which introduced `git branch --show-current`) came out almost 2 years ago. You can change it yourself I guess. Sorry 😞

AdrianTM:
Fair enough thanks for humoring me.

PR #147, opened 2021-04-19
Add default item for Elixir projects

Fibrasek:
#### Description

Add` an default item to the project when navigating Elixir project directories.

#### Motivation and Context

I don´t think there´s a proper problem that this solves, but it´s a nice convenience for the Elixir developers and it broadens the amount of languages with items out of the box.

The method I used to identify an Elixir project is to check the presence of a `mix.exs` file in the directory.

2021-04-19-193111_957x44_scrot

#### How Has This Been Tested

A simple way to test is navigating through multiple projects with different Elixir and OTP versions.

#### Checklist

### DIFF
--- a/functions/_tide_item_elixir.fish
+++ b/functions/_tide_item_elixir.fish
@@ -1,0 +1,10 @@
+function _tide_item_elixir
+    mix deps >/dev/null 2>/dev/null
+
+    if test $status = 0
+        set -l elixirVersion (elixir -v | sed -n 's/Elixir //p;s/compiled with Erlang\///p')
+
+        set_color $tide_elixir_color
+        printf '%s' $tide_elixir_icon ' ' $elixirVersion
+    end
+end
--- a/functions/_tide_item_elixir.fish
+++ b/functions/_tide_item_elixir.fish
@@ -1,6 +1,6 @@
 function _tide_item_elixir
     if test -e mix.exs
-        set -l elixirVersion (elixir -v | sed -nr 's/Elixir //;s/compiled with Erlang\///p')
+        set -l elixirVersion (elixir -v | sed -n 's/Elixir //;s/compiled with Erlang\///p')
 
         set_color $tide_elixir_color
         printf '%s' $tide_elixir_icon ' ' $elixirVersion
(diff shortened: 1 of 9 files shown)


IlanCosman:
Thanks for the PR 😄 There are a few problems, though it's my fault for not addressing them in the contributing docs:

1. As you saw the macOS tests failed initially because of `sed` differences on bsd and linux. That's why we always use Fish builtins such as `string match`. Fish builtins also have much better performance. You can take a look at the `php` item and the Fish docs to get a better understanding of `string match`.
2. You have to add the item and its variables to each default config. Add `elixir` to the end of each `tide_left_prompt_items`. Also, define default values for `tide_elixir_icon` and `tide_elixir_color` in each config. I'll make this better for v5 I promise 😄
3. Add your item to the list of items in `_tide_remove_unusable_items`. This makes it so that the item will be removed if `elixir` cannot be found.

Also, question: I know nothing about Elixir, so what's the value in knowing the "(compiled with Erlang/OTP 23)" part?

Fibrasek:
@IlanCosman that´s my first time doing something in the `fish` environment hehe

It´s great to know the `string match` detail, I´ll see the docs to understand how it works and the other points reflects that I need to read more Tide source :D

It´s important to know the OTP version since it reflects the Erlang environment that you´re working on since Elixir is compiled to work on top BEAM, the Erlang VM :)

IlanCosman:
Unfortunately I've discovered an odd elixir/fish bug which makes this not a possibility for Tide.

`echo (fish --command "elixir --version" &;sleep .01)` prints an error: 
```
init terminating in do_boot ({terminated,[{io,put_chars,[standard_io,unicode,[_]],[]},{Elixir.Kernel.CLI,parse_shared,2,[{_},{_}]},{Elixir.Kernel.CLI,shared_option?,3,[{_},{_}]},{Elixir.Kernel.CLI,mai

Crash dump is being written to: erl_crash.dump...done
```

`echo (fish --command "elixir --version" &;sleep .3)` works though.

I only just now actually tried out the item because it looked just fine, and it prints this error. Sorry 😞

PR #149, opened 2021-05-04
fix: commandline: Can not set commandline in non-interactive mode

maxneuds:
#### Description

I don't use the most recent version since my PopOS only has version 3.1.2 of fish but the problem might also be in the current master.
It happend when I added the conda hook to fish, I got the error message:

```

~/.config/fish/functions/fish_prompt.fish (line 33):
```

The error message is straight forward and there is an easy fix by wrapping the code block into an if clause.
I did this to the current master. Would be nice if this could be tested if it's still necessary. 
Cheers :)


IlanCosman:
This bug was fixed in version 4.2.0, but thanks for taking initiative 😄

I highly recommend using the latest version of Fish from the [official PPA](https://launchpad.net/~fish-shell/+archive/ubuntu/release-3).

PR #170, opened 2021-07-08
Handle pipenv names

WillJBrown:
Feat: Handle Pipenv Virtual Environment Names

This gets current directory name and compares with pipenv's
naming pattern. If it matches the output is changed to the
directory name instead or the virtual environment name.

#### Motivation and Context

Pipenv automatically generates and controls virtual environment names.
They take the form VenvRoot-######### where ######### represents a hash relating to the virtual environment.
This is unsightly and unnecessary - especially where horizontal space is at a premium.
I have reworked it so it just shows the name of the Virtual environment root directory.
I was also considering just making it show 'pipenv' but considering the comment in the existing code about not displaying generics I decided to go with the root directory name.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_virtual_env.fish
+++ b/functions/_tide_item_virtual_env.fish
@@ -1,13 +1,20 @@
 function _tide_item_virtual_env
     if set -l splitVirtualEnv (string split '/' "$VIRTUAL_ENV")
         set_color $tide_virtual_env_color
-        set -l directoryName (string split '/' (pwd))[-1]
-        string match -qe $directoryName- $splitVirtualEnv[-1]; set pipenvStyle $status
+        set -l pathSplit (string split '/' (pwd))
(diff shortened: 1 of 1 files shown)


IlanCosman:
Hi Will, thanks for contributing 😄

When is it the case that the venvroot isn't the last directory? I.e when does `$VIRTUAL_ENV` not look like
`/home/ilan/.local/share/virtualenvs/python-project-EwRYuc3l`?

WillJBrown:
Hi,
It took me a while to see it bu that is indeed a simpler test of if we are in a pipenv environment or not. Thanks.
New commit is much cleaner.

IlanCosman:
Merged your idea into the v5 branch with the above commit. Thanks for contributing @WillJBrown 😄

PR #174, opened 2021-07-12
Add tide prompt item GO similiar to RUST support

dr-emacs:
#### Description

Added prompt item for GO indicating go version number and small gopher icon

#### How Has This Been Tested

- [ X ] I have tested using **Linux**.

#### Checklist

- [ X ] I have updated the tests accordingly.

### DIFF
--- a/functions/_tide_item_go.fish
+++ b/functions/_tide_item_go.fish
@@ -1,6 +1,6 @@
 function _tide_item_go
     if test -e go.mod
         set_color $tide_go_color
-        printf '%s' $tide_go_icon' ' (go version | string split ' ' | string trim --chars go)[3]
+        printf '%s' $tide_go_icon' ' (go version | string split ' ')[3]
     end
 end
--- a/functions/_tide_item_go.fish
+++ b/functions/_tide_item_go.fish
@@ -1,0 +1,6 @@
+function _tide_item_go
+    if test -e go.mod
+        set_color $tide_go_color
+        printf '%s' $tide_go_icon' ' (go version | string split ' ' | string trim --chars go)[3]
+    end
+end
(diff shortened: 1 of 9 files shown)


IlanCosman:
Great work 👍 Merged into the `v5` branch with a few small tweaks. Thanks for contributing 😄

dr-emacs:
Thank you! My first pull reuest ...
When exchanging the gopher icon in my second commit I noticed a very strange behaviour. Somehow the writing direction seems to switch to left to right. At least this improved only after switching from  a numeric only 1.16.5 to go1.16.5.
Maybe you have got en idea?

I'm currently planning to implement a prompt item showing the current battery level. Would that be interesting too?

IlanCosman:
> When exchanging the gopher icon in my second commit I noticed a very strange behaviour. Somehow the writing direction seems to switch to left to right. At least this improved only after switching from a numeric only 1.16.5 to go1.16.5.

I noticed that too, so I just used a different gopher icon 😂 No idea why that specific character is weird.

> I'm currently planning to implement a prompt item showing the current battery level. Would that be interesting too?

Sure! I think it might be a difficult though with different operating systems and whatnot 😄

dr-emacs:
I am only able to test on Linux. I think I'll have to check for the linux specific /sys/class/power_supply. Maybe somebody else will be able to pick up from there ...
Are there more  missing/needed prompt items?
I switched from ksh to bash to zsh to fish and used by current IP, battery level, free memory, CPU temperature, system load.

IlanCosman:
If you're really interested in slaving away 😄 then [Spacefish](https://github.com/matchai/spacefish/tree/master/functions) has a bunch of items that might be nice to port over. Specifically, kubectl has been requested for Tide in the past. 

As for the ones you mentioned, I don't find them useful, and they're probably fairly annoying to implement due to the aforementioned OS issue, but if you want to, go ahead 😄

PR #178, opened 2021-07-17
add kube context

branchvincent:
#### Description

Adds a `kube_context` item for display the current kubernetes cluster/namespace.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_kube_context.fish
+++ b/functions/_tide_item_kube_context.fish
@@ -1,4 +1,4 @@
 function _tide_item_kube_context
     set -l context (kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null)
-    test -n "$context" && _tide_print_item kube_context $tide_kube_context_icon' ' (string replace -r '/default$' '' $context)
+    test -n "$context" && _tide_print_item kube_context $tide_kube_context_icon' ' (string replace -r '/(?:default)?$' '' $context)
 end
--- a/functions/_tide_item_kube_context.fish
+++ b/functions/_tide_item_kube_context.fish
@@ -1,0 +1,4 @@
+function _tide_item_kube_context
+    set -l context (kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null)
+    test -n "$context" && _tide_print_item kube_context $tide_kube_context_icon' ' (string replace -r '/default$' '' $context)
+end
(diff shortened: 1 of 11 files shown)


IlanCosman:
Thanks for working on this, I know it's been requested a number of times in the past 👍

A few things:

1. Could you give me a run down on how to demonstrate the item? Idk kubernetes at all 😄

2. Colors in `lean`, `rainbow`, and `classic` should be in hex format. The point is that they'll look the same anywhere, in contrast with 16_color configs. Don't worry about selecting a hex color though, I can do that 😄

3. Spitballing: Why shouldn't we call the item `kubectl`?

branchvincent:
> 1. Could you give me a run down on how to demonstrate the item? Idk kubernetes at all 😄

Sure! For some background on `kubectl` and its `context` I'll quote the [docs](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context): 

> A context element in a kubeconfig file is used to group access parameters under a convenient name. Each context has three parameters: cluster, namespace, and user. By default, the kubectl command-line tool uses parameters from the current context to communicate with the cluster.

So, `kubectl` interacts on behalf of a `user` with a single Kubernetes `cluster` and a single `namespace` within that cluster. But how do you manage multiple clusters? That's why the context exists and why it's valuable to know at-a-glance what `kubectl` will be interacting with before you do it.

You can play around with this yourself without having a real cluster. All you need is `kubectl` and a valid `kubeconfig` file (which is just yaml). Assuming you use `brew`:

```shell
# Setup
brew install kubernetes-cli
set -a tide_right_prompt_items kube_context

# Create and manipulate the current context
kubectl config set-context my-ctx --namespace default && kubectl config use-context my-ctx # * my-ctx
kubectl config set-context --current --namespace demo-ns1 # * my-ctx/demo-ns1
kubectl config set-context example --namespace default && kubectl config use-context example # * example
kubectl config set-context --current --namespace demo-ns2 # * example/demo-ns2
kubectl config unset current-context # 
cat ~/.kube/config # see the underlying file that stores the above access parameters
```

You can find a more realistic `kubeconfig` and more examples from the [docs](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#define-clusters-users-and-contexts).

> 2. Colors in `lean`, `rainbow`, and `classic` should be in hex format. The point is that they'll look the same anywhere, in contrast with 16_color configs. Don't worry about selecting a hex color though, I can do that 😄

😅 lol thanks for the explanation, i was just guessing

> 3. Spitballing: Why shouldn't we call the item `kubectl`?

good question, i copied [powerlevel10k](https://github.com/romkatv/powerlevel10k#batteries-included) which calls it `kubecontext` (as did [spacefish](https://github.com/matchai/spacefish/blob/master/functions/__sf_section_kubecontext.fish) but now its successor [starship](https://starship.rs/config/#kubernetes) calls it `kubernetes`). i do like the symmetry of calling it `kubectl`. i'm 👍 with whatever you think

IlanCosman:
Thank you for explaining how I can demonstrate locally so thoroughly 👍  Made some changes you can peruse if you wish 😄

Thanks again for doing this, I wasn't going to do it on my own 😂

PR #181, opened 2021-07-24
add node

branchvincent:
#### Description

Adds a `node` item. Note this does overlap with `nvm`, but not everyone uses the latter.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_node.fish
+++ b/functions/_tide_item_node.fish
@@ -1,0 +1,3 @@
+function _tide_item_node
+    test -e package.json && _tide_print_item node $tide_node_icon' ' (node --version | string match --regex "[\d.]+")
+end
--- a/functions/_tide_item_node.fish
+++ b/functions/_tide_item_node.fish
@@ -1,3 +1,3 @@
 function _tide_item_node
-    test -e package.json && _tide_print_item node $tide_node_icon' ' (node --version | string match --regex "[\d.]+")
+    test -e package.json && _tide_print_item node $tide_node_icon' ' (node --version | string trim --chars=v)
 end
--- a/functions/_tide_remove_unusable_items.fish
+++ b/functions/_tide_remove_unusable_items.fish
@@ -1,6 +1,6 @@
 function _tide_remove_unusable_items
     # Remove tool-specific items for tools the machine doesn't have installed
-    for item in chruby git go kubectl nvm php rustc virtual_env
+    for item in chruby git go kubectl node nvm php rustc virtual_env
         set -l cli_names $item
         switch $item
             case virtual_env
--- a/functions/tide/configure/configs/classic.fish
+++ b/functions/tide/configure/configs/classic.fish
@@ -45,6 +45,9 @@ tide_left_prompt_prefix ''
 tide_left_prompt_separator_diff_color ''
 tide_left_prompt_separator_same_color ''
 tide_left_prompt_suffix ''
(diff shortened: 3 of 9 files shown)


IlanCosman:
Seems good 👍 My biggest concern is that it might not be useful to most people. I know that the few times I've worked on a node project, I haven't cared what version I was working with, but perhaps that's abnormal. 

I'm also concerned about the overlap with the nvm item. Granted, not everyone uses nvm, but for those who do this will end up showing the node version twice, which is silly.

For these reasons, I'd suggest having this be off by default, like [powerlevel10k does](https://github.com/romkatv/powerlevel10k/blob/e3c8529052bb8c73fdde48a28ed9f6a8b4a4408b/config/p10k-lean.zsh#L51-L54). I'm open to being convinced otherwise though.

kidonng:
I would suggest checking for more files/folders beside `package.json`, you can refer to [Starship's default config](https://github.com/starship/starship/blob/80c5eca2ee3969f24d74fd1f169228af5c8ca2b2/src/configs/nodejs.rs#L28-L30).

In my own implementation I'm checking for `test -f package.json || test (count *.{{,c,m}j,t}s) != 0`, if it helps.

> My biggest concern is that it might not be useful to most people.

I'd like to agree with you, but I enable this all the time so I can't decide 😄

> I'm also concerned about the overlap with the nvm item. Granted, not everyone uses nvm, but for those who do this will end up showing the node version twice, which is silly.

Starship (again) doesn't have a nvm item but only a Node item. IMO that suits most people and is compatible with (any) nvm. Last but not least, another reason to enable the item all the time!

branchvincent:
> My biggest concern is that it might not be useful to most people.

Personally, I find it helpful since I work on a bunch of different microservices, each with a slightly different node version. But I'll be happy as long as I (and others) can enable it 😄 

> I would suggest checking for more files/folders beside package.json

👍 sounds good to me, I just copied [powerlevel10k](https://github.com/romkatv/powerlevel10k/blob/8f798f986a02eb27c4ef25229976443fa2a66809/internal/p10k.zsh#L2381). Not to scope creep here, but it would be great if this was eventually configurable.

> IMO that suits most people and is compatible with (any) nvm.

Completely agree, I think it makes more sense to have one language version item rather than an item for every single version manager out there. For example, I use [fnm](https://github.com/Schniz/fnm) for node but all I really care about is the output of `node --version` (of course, assuming that my version manager is properly configured).

IlanCosman:
Sorry for the delay, had to finish Finals 😅

Alright, yeah, I agree with you guys. So we remove `nvm` and add `node`.

> I would suggest checking for more files/folders beside package.json... In my own implementation I'm checking for test -f package.json || test (count *.{{,c,m}j,t}s) != 0, if it helps.

Can I ask why? AFAIK basically every node project has a package.json. 

And I don't want to do file globbing, mostly because it's performance sucks. `time begin; count *.js; end` takes 300-400 microseconds on my new Mac. Yah that's not the end of the world, but it takes a very outsized amount of time for the value it adds, which as I said above I'm not sure is any at all.

> Not to scope creep here, but it would be great if this was eventually configurable.

Configurable functionality isn't the Fish way, it's harder to maintain, and I don't really see much benefit here.

P.S. @kidonng `test (count *.{{,c,m}j,t}s) != 0` can be simplified into `count *.{{,c,m}j,t}s >/dev/null` for a little speed boost.

PR #182, opened 2021-07-25
Use single underscore for `__tide_on_fish_exit`

kidonng:
#### Description

Use single underscore for `__tide_on_fish_exit`, by replicating its functionality directly in `_tide_init_uninstall`, which is also more suitable IMO.

#### Motivation and Context

Recently I've been thinking that most Fish plugins agree on using a single underscore for their "private" functions, while Fish exclusively uses two (except for `_validate_int`, which is being tackled with fish-shell/fish-shell#8168). Tide also follows this convention except for `__tide_on_fish_exit` due to technical reason, but I just realized this can solved the other way.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/conf.d/_tide_init.fish
+++ b/conf.d/_tide_init.fish
@@ -17,6 +17,6 @@ function _tide_init_install --on-event _tide_init_install
 end
 
 function _tide_init_uninstall --on-event _tide_init_uninstall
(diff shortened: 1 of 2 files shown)


IlanCosman:
Seems good 👍

PR #190, opened 2021-08-27
Make Git icons configurable

radek-sprta:
Makes git prompt icons configurable through variables.

#### Description

Adds the following variables:
* `tide_git_icon_ahead '⇡'`
* `tide_git_icon_behind '⇣'`
* `tide_git_icon_conflicted '~'`
* `tide_git_icon_dirty '!'`
* `tide_git_icon_staged '+'`
* `tide_git_icon_stash '*'`
* `tide_git_icon_untracked '?'`

#### How Has This Been Tested

I use Tide with this patch and custom icons. I didn't see any problem.

#### Checklist

**I'll update the wiki, if merged.**
**Other prompt items do not test icon configurability either.**

### DIFF
--- a/functions/_tide_item_git.fish
+++ b/functions/_tide_item_git.fish
@@ -62,10 +62,10 @@ function _tide_item_git
 
     _tide_print_item git $_tide_location_color $tide_git_icon' ' (set_color white; printf %s $location
         set_color $tide_git_color_operation; printf %s ' '$tide_git_operation ' '$tide_git_step/$tide_git_total_steps
-        set_color $tide_git_color_upstream; printf %s ' ⇣'$upstream_behind ' ⇡'$upstream_ahead
-        set_color $tide_git_color_stash; printf %s ' *'$stash
-        set_color $tide_git_color_conflicted; printf %s ' ~'$conflicted
-        set_color $tide_git_color_staged; printf %s ' +'$staged
-        set_color $tide_git_color_dirty; printf %s ' !'$dirty
-        set_color $tide_git_color_untracked; printf %s ' ?'$untracked)
+        set_color $tide_git_color_upstream; printf %s ' '$tide_git_icon_behind$upstream_behind ' '$tide_git_icon_ahead$upstream_ahead
+        set_color $tide_git_color_stash; printf %s ' '$tide_git_icon_stash$stash
+        set_color $tide_git_color_conflicted; printf %s ' '$tide_git_icon_conflicted$conflicted
+        set_color $tide_git_color_staged; printf %s ' '$tide_git_icon_staged$staged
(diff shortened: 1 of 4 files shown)


IlanCosman:
In general I'm leaning towards being less configurable than we currently are, but I'll think more about it 👍 As I said in #191, I'm on vacation so I'll address this when I get back.

radek-sprta:
I realize each configuration options comes with maintenance cost. However, for configurable icons and colors, it does not seem very high.

IlanCosman:
Sorry for the delay on this. I'm hesitant to merge here, as I've been seriously considering _removing_ most icon configurability.

My thought process is like this:

0. Configurable icons = more complexity. Not by a lot, but they still have a cost for documentation, and reduce "cleanliness" for the user. The user shouldn't have any more control than they need. And they don't need this control because:
1. The default icons are essentially perfect.
    - And if someone can make a good argument for why there are better icons, then _those_ should become the default icons.
2. If the default icons are basically perfect, then the user changing them would be making them worse, or at least not better.
3. Therefore, the only reason someone would change their icons is that they are accustomed to other icons from a different prompt.

And that's what I don't like. I don't like making Tide more complex, exposing more to the user, and letting people downgrade/sidegrade their icons, just so that they can recreate something from a different prompt.

I know, this is quite some melodrama over icons 😄 Thoughts?

radek-sprta:
No worries.

You have a good point in that the default icons are perfect, because however plain, they are actually the icons that git status uses. And changing them wouldn't even be a good idea, because Tide users are used to these. Ultimately, making them configurable, makes the transition to Tide easier for users, who are used to a different set of icons. Changing habits is hard and getting used to a new set of icons might actually lower person's productivity for several weeks.

I guess it also depends how far you want to go in removing the icon configuration. If you want to keep at least the option of many icons vs no icons, you still need the icon variables. And even if you go as far as removing the icon configuration altogether, I would argue, that code would be more readable having the icons defined as variables along the colors, rather than hard-coded in each prompt segment.

When it comes to documentation, since has a special section in wiki and is not part of the readme, I don't think makes Tide more complex for the average user. Only the people actively looking for this information will come upon it.

All that said, you are the maintainer and I don't want to pressure you into merging something you don't want to. So if you want to go in the direction of less configurability, feel free to close this PR.

IlanCosman:
> they are actually the icons that git status uses

Side-note: What do you mean by this?

radek-sprta:
> > they are actually the icons that git status uses
> 
> Sidenote: What do you mean by this?

Just my mind playing tricks on me :) I thought those were the symbols used by `git status --porcelain`, but that's only true for untracked files.

IlanCosman:
Sorry for ignoring this. Yes, having the git icons be unconfigurable is inconsistent, but bah, I feel like doing it anyway. I'll think about it more though.

PR #192, opened 2021-08-30
add ws integration

daveio:
Adds support for displaying the current `ws` workspace. `ws` is a flexible workspace switcher which can be configured to run various commands to switch between different environments. Think Kubernetes contexts but for the entire shell. See https://github.com/daveio/ws for more.

#### Description

Add `_tide_item_ws` and `tide_ws_icon`, add `ws` item to configs on the right side next to `node`.


daveio:
Apologies for the commit train, this is now ready for review. Tested on macOS.

IlanCosman:
Hi Dave, thanks for taking the time to contribute 😄 

Unfortunately, I'd like to wait on adding support for ws until it gets more popular. While it may not have a performance impact if the item isn't used, any new item has a code/documentation cost.

I hope your project does well 👍

daveio:
Seems fair enough to me. I'll keep using my fork, just thought I'd pop you a PR just in case :)

PR #203, opened 2021-09-18
Add Terraform item

radek-sprta:
#### Description

Adds a new item - terraform - to show the terraform workspace. The colors are provisional, I can change them to something different. Terraform does not have an icon in nerd fonts, so I left it undefined.

#### Motivation and Context

For people who work a lot with Terraform, seeing the current workspace may prevent costly mistakes.

#### How Has This Been Tested

Tried with terraform on my installation. Correctly shows current workspace when inside terraform directory and nothing outside of it. I also added unit tests.

#### Checklist

**I will update the wiki if changes are merged**
- [X] I have updated the tests accordingly.

### DIFF
--- a/functions/_tide_item_terraform.fish
+++ b/functions/_tide_item_terraform.fish
@@ -1,0 +1,3 @@
+function _tide_item_terraform
+    test -d .terraform && _tide_print_item terraform $tide_terraform_icon' ' (terraform workspace show)
+end
--- a/functions/_tide_item_terraform.fish
+++ b/functions/_tide_item_terraform.fish
@@ -1,6 +1,6 @@
 function _tide_item_terraform
     if test -d .terraform
-        set -l workspace (terraform workspace show)
-        test $workspace != 'default' && _tide_print_item terraform $tide_terraform_icon' ' $workspace
+        terraform workspace show | read -l workspace
+        test $workspace != default && _tide_print_item terraform $tide_terraform_icon' ' $workspace
     end
 end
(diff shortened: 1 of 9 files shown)


IlanCosman:
A few things:

1. Can you describe how to test this in reality? I.e what commands do I run to set up a simple terraform workspace?
2. We probably don't want to display this when it just says `default`. 
3. Powerlevel10k uses `$TF_WORKSPACE` [to do this task](https://github.com/romkatv/powerlevel10k/blob/4f3d2ffe7251a17aa10e00ed30d55f6fd9d65002/internal/p10k.zsh#L4884-L4888). Can we use that instead of `(terraform workspace show)`? It would be much faster because we wouldn't be interacting with terraform at all.

radek-sprta:
> A few things:
> 
>     1. Can you describe how to test this in reality? I.e what commands do I run to set up a simple terraform workspace?
> 
You can quickly test it like this:

```
cd (mktemp -d)
terraform init
terraform workspace new test
terraform workspace select default
```

>     2. We probably don't want to display this when it just says `default`.
Good point.
 
>     3. Powerlevel10k uses `$TF_WORKSPACE` [to do this task](https://github.com/romkatv/powerlevel10k/blob/4f3d2ffe7251a17aa10e00ed30d55f6fd9d65002/internal/p10k.zsh#L4884-L4888). Can we use that instead of `(terraform workspace show)`? It would be much faster because we wouldn't be interacting with terraform at all.
`$TF_WORKSPACE` doesn't always exist - it's an override. Normally, the current workspace is in `.terraform/environment` (and the directory itself can be overriden with `TF_DATADIR`).
I have chosen to use `terraform workspace show` because it is future-proof, should anything change, and makes the code cleaner, while remaining fast. But if you want me to rewrite it using env variables and `cat`, let me know.

PR #217, opened 2021-10-28
Add gcloud item for Google Cloud Platform

maxekman:
#### Description

Adds a  gcloud item to show the current Google Cloud Platform project.

#### Motivation and Context

I have previously showed the current GCP project in the tmux tab bar but displaying it as an item in the command line reduces the chance of doing operations to the wrong project considerable.

Screenshot 2021-10-28 at 17 43 21

#### How Has This Been Tested

Tested using a real GCP project and the project unit testing.

#### Checklist

### DIFF
--- a/functions/_tide_item_gcloud.fish
+++ b/functions/_tide_item_gcloud.fish
@@ -1,0 +1,4 @@
+function _tide_item_gcloud
+    set -l project (gcloud config get-value core/project 2>/dev/null) &&
+        _tide_print_item gcloud $tide_gcloud_icon' ' $project
+end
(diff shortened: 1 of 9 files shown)


IlanCosman:
> I could try to read the config files directly, but that would possibly lead to breakage or not support all types of installs

Yah, if you can find some way to make it faster while also portable and stable, that'd be amazing, but I don't expect that's possible. Until then, I'm going to close this PR.

I really appreciate you contributing though, even if it didn't work out 😄

maxekman:
Understandable, I’ll reopen it if I figure it out. Thanks for reviewing.

dudeofawesome:
What is the target latency for a prompt item?

IlanCosman:
@dudeofawesome It depends on how important the item is, and how often we expect it to run. E.g. git and pwd are extremely important and therefore have a much higher threshold. Some other item might be pretty slow, but if it only runs when a certain file is present, then it's more acceptable.

In general though 100ms is an upper limit (of course this depends on your machine as well).

PR #219, opened 2021-10-30
feat: Add context support for toolbox

aidenlangley:
#### Description
Adding a context check - if `test -f /run/.toolboxenv` passes, we are in a toolbox, so display `hostname` in `context`. User remains the same.

#### Motivation and Context

Small quality of life change - when using `toolbox` it can be hard to identify where you are, because a `toolbox` shares your /home/ and a lot of host information, but can be treated like a somewhat isolated container by the user.

#### How Has This Been Tested

Tested by entering and exiting a `toolbox` to see if `context` would appear & disappear.

#### Checklist

### DIFF
--- a/functions/_tide_item_context.fish
+++ b/functions/_tide_item_context.fish
@@ -3,6 +3,8 @@ function _tide_item_context
         tide_context_color=$tide_context_color_ssh _tide_print_item context $USER'@'$hostname
     else if contains -- $USER root toor Administrator
         tide_context_color=$tide_context_color_root _tide_print_item context $USER'@'$hostname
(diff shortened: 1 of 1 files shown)


IlanCosman:
A toolbox item was added in https://github.com/IlanCosman/tide/commit/38466cf6a110fe9ff66ec0d98385382639d0a3d3 (will be released with v5.1) Perhaps I should add `$hostname` to it as well though.

I prefer to have toolbox as a separate item because the vast majority of users won't have toolbox installed.

IlanCosman:
Thanks for contributing though 😄

aidenlangley:
Looking forward to the release. ^^

PR #221, opened 2021-10-31
Add truncation possibility for git prompt

svenihoney:
#### Description

In case of looong git branch names the prompt fills up the whole screen. This PR introduces variables `tide_git_truncation_length` and `tide_git_truncation_symbol` (predefined to "…") to truncate the git prompt after the given number of characters.

#### How Has This Been Tested

1. Created a long branch name
2. Observed the prompt

#### Checklist

- [ ] I have updated the documentation accordingly. Variable doumentation is in the Wiki...

### DIFF
--- a/conf.d/_tide_init.fish
+++ b/conf.d/_tide_init.fish
@@ -2,6 +2,7 @@ function _tide_init_install --on-event _tide_init_install
     set -U tide_os_icon (_tide_detect_os)
     set -U VIRTUAL_ENV_DISABLE_PROMPT true
     set -U _tide_var_list tide_os_icon VIRTUAL_ENV_DISABLE_PROMPT
+    set -U tide_git_truncation_symbol "…"
 
     source (functions --details _tide_sub_configure)
     _load_config lean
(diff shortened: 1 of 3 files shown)


deflock:
What do you think about adding also variable `truncation_position` with values `start`, `middle`, `end`?

IlanCosman:
I'm not a fan of configurable functionality @deflock. Also, `middle` would probably be fairly difficult to implement.

IlanCosman:
Thank you for the idea @svenihoney, and thank you for contributing 😄
I found a solution I like a bit more which uses `string replace`.

svenihoney:
> Thank you for the idea @svenihoney, and thank you for contributing smile I found a solution I like a bit more which uses `string replace`.

I have seen your solution, but it means that you have to set tide_git_truncation_length in any case or it gives an error. Okay, you have a workaround by setting length to a high value, which should be the default instead of 24 IMHO.

IlanCosman:
I'm not sure I understand what you're saying. No errors appear even if `tide_git_truncation_length` is erased, or empty.

svenihoney:
Okay, my bad. Did not test correctly, sorry. Will you backport this to the main branch or will it stay in 3.4 branch only and I have to wait for fish 3.4 to use your upstream again?

IlanCosman:
I won't be backporting anything, sorry. Hopefully Fish 3.4 will release soon.

PR #230, opened 2021-12-02
Check if PHP is installed before trying to show version

ruudk:
When browsing to a directory with `composer.json` without PHP installed, you get this error on every prompt:
```
❯ fish: Unknown command: php
~/.config/fish/functions/_tide_item_php.fish (line 1):
php --version | string match --regex 'PHP ([\d.]+)'
^
in command substitution
	called on line 2 of file ~/.config/fish/functions/_tide_item_php.fish
in function '_tide_item_php'
	called on line 34 of file ~/.config/fish/functions/_tide_prompt.fish
in function '_tide_left_prompt'
	called on line 1 of file ~/.config/fish/functions/_tide_prompt.fish
in command substitution
	called on line 2 of file ~/.config/fish/functions/_tide_prompt.fish
in function '_tide_prompt'
in command substitution
~/.config/fish/functions/_tide_item_php.fish (line 2): Unknown command
    test -e composer.json && _tide_print_item php $tide_php_icon' ' (php --version | string match --regex 'PHP ([\d.]+)')[2]
                                                                    ^
in function '_tide_item_php'
	called on line 34 of file ~/.config/fish/functions/_tide_prompt.fish
in function '_tide_left_prompt'
	called on line 1 of file ~/.config/fish/functions/_tide_prompt.fish
in command substitution
	called on line 2 of file ~/.config/fish/functions/_tide_prompt.fish
in function '_tide_prompt'
in command substitution
```

This should fix that.

I think this same principle should be applied to all other commands too. What do you think?

### DIFF
--- a/functions/_tide_item_php.fish
+++ b/functions/_tide_item_php.fish
@@ -1,3 +1,3 @@
 function _tide_item_php
-    test -e composer.json && _tide_print_item php $tide_php_icon' ' (php --version | string match --regex 'PHP ([\d.]+)')[2]
+    test -e composer.json && which php && _tide_print_item php $tide_php_icon' ' (php --version | string match --regex 'PHP ([\d.]+)')[2]
 end


IlanCosman:
There's [a system in place](https://github.com/IlanCosman/tide/blob/main/functions/_tide_remove_unusable_items.fish) to prevent this type of error already. If you restart fish, this shouldn't happen again (unless you're adding `php` to the items list manually). 

Regardless, I've improved the system for v5.1 already, so people hopefully shouldn't run into this problem in the future, even if it can be solved by a terminal restart. Thanks for being proactive about your problem though 👍

ruudk:
I think this happened because I removed `php` while keeping my terminal open.

I think it should just handle this, and not print a long stacktrace.

Another thing that I noticed, is that when I change the `PATH` in the terminal, so that another version of `php` gets precedence, the showed version does not update. 

It always stays the version that it was started with.

IlanCosman:
> I think this happened because I removed php while keeping my terminal open.

That would do it 😄

> I think it should just handle this, and not print a long stacktrace.

Unfortunately there's a tradeoff here. I could put guards in each item, but that would slow things down a bit. Not only would it have to run `type -q` (builtin version of which), but it would have to load the item function in the first place. This way nothing gets loaded that isn't needed.

> Another thing that I noticed, is that when I change the PATH in the terminal, so that another version of php gets precedence, the showed version does not update.

That's because Tide runs everything in a separate fish process. You'd have to export things like that for them to affect tide.

Issue #90, opened 2021-02-09
Add powerlevel10k's "Show on command"


Lameorc:
#### Is your feature request related to a problem? Please describe

Certain "info" (eg. `kubectl config current-context` or parts of `gcloud info`) is only relevant when running certain commands.
Currently it's pretty easy to show it thanks to custom items but it is irrelevant in plenty of situations and just clutters up
the prompt. 

#### Describe the solution you'd like

The prompt item is only visible when relevant command is part of the user input.

#### Additional context

[powerlevel10k's section in readme](https://github.com/romkatv/powerlevel10k/#show-on-command)

I'm not sure how feasible this is in fish without some sort of fairly obvious flickering due to the async nature of this.

IlanCosman:
@faho Could you help me out here? Seemed more efficient than gitter 😂 

Reading the commandline buffer and parsing it is easy, but the trigger to do so is the hard part. Got any ideas? Or is this not doable?

faho:
No, not feasible.

branchvincent:
I also really wanted this feature, so I made a hacky attempt to get close. I chose my tigger to just be the user inserting a space:

```fish
function _tide_show_on_command
    if test (count (commandline -poc)) -eq 0
        set -l cmd (commandline -t)
        # Check to see if this is an abbr that will be expanded
        if abbr -q $cmd
            set -l var _fish_abbr_$cmd
            set cmd $$var
        end
        switch $cmd
            case kubectl helm kubens kubectx
                set -gx tide_show_k8s # export this so our item sees it
                commandline -f repaint
            case '*'
                set -e tide_show_k8s
                commandline -f repaint
        end
    end
end

bind ' ' 'commandline -f expand-abbr; _tide_show_on_command; commandline -i " "'
```

and then using it would look like:
```fish
function _tide_item_k8s --description "Show Kubernetes context"
    if set -q tide_show_k8s
        # ...display context 
    end
end
```

Note, this very naive and far from as elegant as powerlevel10k's version, but it's close enough for me. Still, I'm interested in better ways to achieve this and (if feasible) it being officially added to tide!

douglascamata:
Maybe we can break this issue into separate things? I understand that "show on command" is a great feature. But to me it's separate from "showing the k8s context". And the k8s context is super important for many people, it could be given priority if @IlanCosman agrees with it.

[...]

PR #231, opened 2021-12-05
Add docker context item

Crocmagnon:
#### Description
Add an item displaying the current docker context. Largely inspired by the kubectl item.

Screenshot 2021-12-05 at 20 36 24

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_docker.fish
+++ b/functions/_tide_item_docker.fish
@@ -1,5 +1,5 @@
 function _tide_item_docker
-    set -l context (_tide_item_docker_replace_ignore_result (docker context show 2>/dev/null)) &&
+    set -l context (_tide_item_docker_replace_ignore_result (docker context inspect --format "{{.Name}}" 2>/dev/null)) &&
         test -n "$context" &&
         _tide_print_item docker $tide_docker_icon' ' $context
 end
--- a/functions/_tide_item_docker.fish
+++ b/functions/_tide_item_docker.fish
@@ -1,0 +1,10 @@
+function _tide_item_docker
+    set -l context (_tide_item_docker_replace_ignore_result (docker context show 2>/dev/null)) &&
+        test -n "$context" &&
+        _tide_print_item docker $tide_docker_icon' ' $context
+end
+
+function _tide_item_docker_replace_ignore_result
+    string replace --regex '^default$' '' $argv
+    return 0
+end
--- a/functions/_tide_item_docker.fish
+++ b/functions/_tide_item_docker.fish
@@ -1,10 +1,4 @@
 function _tide_item_docker
-    set -l context (_tide_item_docker_replace_ignore_result (docker context inspect --format "{{.Name}}" 2>/dev/null)) &&
-        test -n "$context" &&
-        _tide_print_item docker $tide_docker_icon' ' $context
-end
-
-function _tide_item_docker_replace_ignore_result
-    string replace --regex '^default$' '' $argv
-    return 0
+    docker context inspect --format '{{.Name}}' | read -l context
+    test "$context" != default && _tide_print_item docker $tide_docker_icon' ' $context
 end
--- a/functions/_tide_remove_unusable_items.fish
+++ b/functions/_tide_remove_unusable_items.fish
@@ -1,6 +1,6 @@
 function _tide_remove_unusable_items
     # Remove tool-specific items for tools the machine doesn't have installed
-    for item in chruby git go kubectl node php rustc terraform virtual_env
+    for item in chruby git go kubectl node php rustc terraform virtual_env docker
         set -l cli_names $item
         switch $item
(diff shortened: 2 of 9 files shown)


IlanCosman:
Thanks! I'll take a look. Don't worry about mega-linter.

IlanCosman:
Can you give me a quick run-down on how you tested this? I have everything set up, but I'm not very knowledgeable about docker.

Crocmagnon:
Sure! With docker installed, you can list contexts using `docket context ls`. There should be at least one named `default`. If you use the default context, no output will be shown as with the kubectl item.
You can create a new context with another name and switch to using it, the item should then update.

Here’s Docker’s documentation about contexts including how to create and switch to a new context: https://docs.docker.com/engine/context/working-with-contexts/

IlanCosman:
Found a problem. `docker context show` isn't available with normal docker, it requires docker-compose. You can switch this item to using `docker context ls` or `docker context inspect` at your discretion.

Crocmagnon:
Done 👍🏻

IlanCosman:
Thanks, looks pretty good 👍 I'll get back to this in a few days, I need to study for finals 😄

IlanCosman:
I'm a bit concerned about performance 😞 Even with async, I'm not super excited about 20-40 ms extra (on my machines). 

`kubectl` has garbage performance, but I'm more ok with that because fewer people use it.

Crocmagnon:
Well, considering that docker contexts are kind of a power user feature, and power users are very likely to also have docker compose, and IIRC, compose is now bundled into docker (it’s now available as a subcommand `docker compose` instead of a separate `docker-compose`) ; would it be ok to revert to the previous better performing version of the item (with context show)?

IlanCosman:
> Well, considering that docker contexts are kind of a power user feature, and power users are very likely to also have docker compose, and IIRC, compose is now bundled into docker (it’s now available as a subcommand docker compose instead of a separate docker-compose) ; would it be ok to revert to the previous better performing version of the item (with context show)?

Just tried it out. `docker compose` is only available _after_ you've installed `docker-compose`. So not really bundled together.
And I haven't gotten `docker context show` to work at all, even with docker compose installed. 

"docker context show" on google gives very few results, and no official documentation.

---

I think that we should follow what spacefish and starship are doing here, which is only running the item if some files exist.
From Starship:

detect_files | ["docker-compose.yml", "docker-compose.yaml", "Dockerfile"] | Which filenames should trigger this module (needs only_with_files to be true).
-- | -- | --

Could you add that? Don't make it configurable though.

Crocmagnon:
Yeah, I couldn't find any official documentation on `docker context show`, but I found it in the github issue template on the official compose-cli repository though (https://github.com/docker/compose-cli/issues/new).

Just for the record, I'm currently using version 20.10.11 of Docker Desktop for Mac. The same version on linux doesn't seem to offer `docker context show` indeed.

Anyway, I tried with `docker context ls | grep "*" | cut -f 1 -d '*' | xargs` but it takes about the same time as the previous version (inspect + format).

I don't think limiting this to a repository containing some arbitrary files would be useful. As with k8s contexts, this affects docker commands globally.

IlanCosman:
> I don't think limiting this to a repository containing some arbitrary files would be useful. As with k8s contexts, this affects docker commands globally.

Yes, it effects docker globally, but the question is when are you going to need to see it. 

The optimal solution AFAICT would be https://github.com/IlanCosman/tide/issues/90, where the docker context only shows up if your command line has `docker` in it.

We generally use the files as a proxy for that though. How often are you going to be running `docker` commands when there are no docker related files nearby? (I don't know the answer btw, I don't use docker 😄)

Crocmagnon:
> How often are you going to be running docker commands when there are no docker related files nearby?

That's a very interesting question. Since I mostly use docker compose to run stuff with docker on my machine, I'd say that in this case I would nearly always have a compose or dockerfile nearby.
However, having recently discovered docker contexts, my use case is slightly different: I'm managing a remote server with docker installed and I want to access its docker daemon easily from my machine, to view logs for example. In this case, I don't think I'll have a compose file nearby, and that's when viewing the context would matter the most.

IlanCosman:
> Is the detect_files function already implemented?

Yes, it's called [test](https://fishshell.com/docs/current/cmds/test.html) 😄

> In this case, I don't think I'll have a compose file nearby, and that's when viewing the context would matter the most.

Fair enough. Then let's not detect files, and just have this item not be on by default 👍 (so no changes necessary)

IlanCosman:
5727ca353f70e3a79daf5176946acdd97d1b9410 Merged with some changes into the fish-3.4 branch 👍 Thanks for contributing 😄

IlanCosman:
Oh, dang, it didn't give you credit for any of your commits. I'll try to fix that.

PR #245, opened 2022-01-06
pyenv support

cr-mitmit:
Support for pyenv right prompt

Introduce _tide_item_pyenv.fish which extracts pyenv version python projects if pyenv is installed on machine

#### Motivation and Context

Closes # 

#### How Has This Been Tested

- [ v] I have tested using **MacOS**.

#### Checklist

### DIFF
--- a/functions/_tide_item_pyenv.fish
+++ b/functions/_tide_item_pyenv.fish
@@ -1,10 +1,10 @@
 function _tide_item_pyenv
-	if test -n "$PYENV_VERSION" \
-		-o -f .python-version \
-		-o -f requirements.txt \
-		-o -f pyproject.toml \
-		-o (count *.py) -gt 0
+    if test -n "$PYENV_VERSION" \
+        -o -f .python-version \
+        -o -f requirements.txt \
+        -o -f pyproject.toml \
+        -o (count *.py) -gt 0
         set -l pyenv_status (pyenv version-name 2>/dev/null) 
         _tide_print_item pyenv $tide_pyenv_icon' ' $pyenv_status 
-	end
+    end
 end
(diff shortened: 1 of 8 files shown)


functions/_tide_item_pyenv.fish
```diff
@@ -0,0 +1,10 @@
+function _tide_item_pyenv
+    if test -n "$PYENV_VERSION" \
+            -o -f .python-version \
+            -o -f requirements.txt \
+            -o -f pyproject.toml \
+            -o (count *.py) -gt 0
+        set -l pyenv_status (pyenv version-name 2>/dev/null)
+        _tide_print_item pyenv $tide_pyenv_icon' ' $pyenv_status
+    end
+end
```
jorgebucaran, line 10: Using Fish's `test` builtin like this has a catch. Even if `PYENV_VERSION` is set and not empty, every expression that follows will be evaluated up, which is not ideal.

Therefore, I think the following would be more efficient:

```fish
function _tide_item_pyenv
    test -n "$PYENV_VERSION" ||
        test -f .python-version ||
        test -f requirements.txt ||
        test -f pyproject.toml ||
        test (count *.py) -gt 0 &&
        _tide_print_item pyenv $tide_pyenv_icon' ' (pyenv version-name 2>/dev/null)
end
```

jorgebucaran (changes requested):
@cr-mitmit I have a suggestion for improving the pyenv item. Everything else looks good to me. 👌

functions/_tide_item_pyenv.fish
IlanCosman: I think actually the new `path` builtin would be best for this. But I'd like to combine this with the `venv` item into a new `python` item for Tide v6.

jorgebucaran: Even better. 👍

IlanCosman:
I think the important functionality of this should be done in the new `python` item, if not, let me know :)

PR #268, opened 2022-03-17
Don't omit last suffix in 2-line prompts

GladOSkar:
#### Description

This previously was in `_tide_prompt.fish` line 36 and 51 but apparently not brought over to the new branch?

I think this PR replicates the original behavior of only adding the suffix on the second line if there is at least one item on the line (`add_prefix` being the new `_tide_last_item_was_newline`).

Feel free to let me know if i'm missing something.

#### How Has This Been Tested

tested with 2-line prompts both with and without items on the second line after the `'newline'` item.

#### Checklist

### DIFF
--- a/functions/_tide_2_line_prompt.fish
+++ b/functions/_tide_2_line_prompt.fish
@@ -3,12 +3,20 @@ function _tide_2_line_prompt
     _tide_side=left for item in $_tide_left_items
         _tide_item_$item
     end
+    if not set -e add_prefix
+        set_color $prev_bg_color -b normal
+        echo $tide_left_prompt_suffix
+    end
 
     echo
 
     set -g add_prefix
     _tide_side=right for item in $_tide_right_items
         _tide_item_$item
+    end
+    if not set -e add_prefix
+        set_color $prev_bg_color -b normal
+        echo $tide_right_prompt_suffix
     end
 end
 


GladOSkar:
I can't see why that macOS test fails, don't think it's because of me. Could someone with macOS check that?

Don't think doc/test updates are necessary for this

IlanCosman:
MacOS tests fail because github hasn't updated their runners to have fish 3.4 yet.

Yah the code works. To be honest I'm not entirely sure I want to allow 2 line prompts like this. In general the entire point of a two line prompt is that you always start typing at the same part of the screen. Thus, generally, one just has the `character` item in the second line. Your use case was basically just a fancier, less useful version of the `character` item (doesn't indicate status, or vi mode, or errors).

I suppose it's a pretty small code penalty though, so I should probably just allow it because it can look nice 🤷‍♂️

GladOSkar:
I don't mind running my own fork if you don't agree with it, no worries. Your call

IlanCosman:
Eh, I'll allow it for now, if only because 5.1 is a minor version number and disallowing this would be a breaking change 😄

Thanks for your help 👍

GladOSkar:
I suppose a more "proper" version would be to allow more styling on the `character` item? I tried changing it to use the powerline character but couldn't really make it work

IlanCosman:
Yah I was considering that, but it would involve a pretty significant extension to the `character` item. The `character` item really does have to be special-cased, at least with the architecture I currently have. For example, it should (optionally) not respond to `tide_prompt_pad_items`, and (optionally) not respond to `add_prefix`.

GladOSkar:
I see, it's difficult. I'll leave you to it then. Thanks for tide, it's pretty awesome :)

PR #283, opened 2022-04-21
Support multiple 'default' contexts for Docker

halostatue:
#### Description

If someone is using 'colima' instead of Docker Desktop, the default context is 'colima'. Rather than making two tests with fixed values, add a new configuration variable and use `contains`.

#### Motivation and Context

I use `colima` and don't want the `colima` context to show, because it's equivalent to the `default` context.

However, rather than hardcoding `colima`, I’ve extended this to have a _list_ of excluded contexts.

#### How Has This Been Tested

I tested with `make test`.

#### Checklist

  - There is no explicit documentation for this, so I’m not sure where to put it. Documenting the variable `tide_docker_default_contexts` (which _does_ include `colima`) would be appropriate in some way.

### DIFF
--- a/functions/tide/configure/configs/classic_16color.fish
+++ b/functions/tide/configure/configs/classic_16color.fish
@@ -10,7 +10,6 @@ tide_context_color_root bryellow
 tide_context_color_ssh yellow
 tide_docker_bg_color black
 tide_docker_color blue
-tide_docker_default_contexts default colima
 tide_git_bg_color black
 tide_git_bg_color_unstable black
 tide_git_bg_color_urgent black
--- a/functions/tide/configure/configs/classic_16color.fish
+++ b/functions/tide/configure/configs/classic_16color.fish
@@ -10,6 +10,7 @@ tide_context_color_root bryellow
 tide_context_color_ssh yellow
 tide_docker_bg_color black
 tide_docker_color blue
+tide_docker_default_contexts default colima
 tide_git_bg_color black
 tide_git_bg_color_unstable black
 tide_git_bg_color_urgent black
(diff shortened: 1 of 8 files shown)


IlanCosman:
Thanks for the PR Austin 🙂 I'm certainly willing to not display on `colima`. 

However, I'm debating the value of adding this as a configuration knob. Is this actually something you think people might be likely to change? It's not that I'm particularly averse to this exact knob, but rather if we introduced a knob for every little thing that maybe, someday, one person might want to change, it would get unreasonable rather quickly 😄

halostatue:
I understand the reluctance to make a configuration knob. However, in this case, I think that it’s necessary. I’m currently using Colima, but I want to play with podman and rancher desktop, which may create / use _different_ contexts. As we discover these, we can add another context to exclude by default (if necessary), but that starts to get a little silly (IMO).

functions/tide/configure/configs/classic_16color.fish
```diff
@@ -10,6 +10,7 @@ tide_context_color_root bryellow
 tide_context_color_ssh yellow
 tide_docker_bg_color black
 tide_docker_color blue
+tide_docker_default_contexts default colima
```
IlanCosman, line 13: This variable shouldn't be in any of the 16_color configs. Those configs are for modifying colors only, they inherit all of the data variables.

halostatue: OK. I’ll adjust that in a few hours and push up a new version.

halostatue: Sorry. Just resolved this.

IlanCosman:
Thanks for the contribution 🙂

PR #293, opened 2022-04-24
Add aws item

okravtsov:
#### Description

Add a new item `aws` to show current value of some environment variables, used to configure aws-cli. These config parameters are current cli profile and region.

#### Motivation and Context

In my job I have to work with many AWS accounts and many AWS regions. I have several pre-configured aws-cli profiles to be able to quick switch between accounts. With **AWS_PROFILE** environment variable I can quick choose right profile to use in current work session. The `aws` item helps to understand at a glance which exactly aws-cli profile are using in exactly terminal tab.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_aws.fish
+++ b/functions/_tide_item_aws.fish
@@ -1,14 +1,11 @@
 function _tide_item_aws
-    # AWS_PROFILE overrides AWS_DEFAULT_PROFILE
-    set -q AWS_DEFAULT_PROFILE && set -f _profile_name $AWS_DEFAULT_PROFILE
-    set -q AWS_PROFILE && set -f _profile_name $AWS_PROFILE
-    # AWS_REGION overrides AWS_DEFAULT_REGION
-    set -q AWS_DEFAULT_REGION && set -f _region_name $AWS_DEFAULT_REGION
-    set -q AWS_REGION && set -f _region_name $AWS_REGION
+    # AWS_PROFILE overrides AWS_DEFAULT_PROFILE, AWS_REGION overrides AWS_DEFAULT_REGION
+    set -q AWS_PROFILE && set -l AWS_DEFAULT_PROFILE $AWS_PROFILE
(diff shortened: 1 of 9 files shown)


IlanCosman:
Thanks for the PR Oleksandr 🙂
Could you give me instructions for setting up a simple aws profile so I can test this?

okravtsov:
@IlanCosman , to test the item you don't have to have real aws-cli profile. Everything that you need is environment variable - _AWS_PROFILE_.
But if you want to test something real you need _access key_ & _secret key_ of 2 different AWS Users, even from one account.
Full test steps:

1. Prepare access keys ID and secret access keys for 2 different AWS users: <https://docs.aws.amazon.com/cli/latest/userguide/getting-started-prereqs.html>
2. Install aws-cli: <https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html>
3. Create 1st profile:
    ```sh
    aws configure --profile profile1
    ```
4. Create 2nd profile:
    ```sh
    aws configure --profile profile2
    ```
5. Set value of _AWS_PROFILE_ to profile1 and test it.
    ```sh
    set -x AWS_PROFILE profile1
    # The item will show 'a profile1'
    aws sts get-caller-identity
    # The output will show you that this is the 1st user
    ```
6. Set value of _AWS_PROFILE_ to profile2 and test it.
    ```sh
    set -x AWS_PROFILE profile2
    # The item will show 'a profile2'
    aws sts get-caller-identity
    # The output will show you that this is the 2nd user
    ```

IlanCosman:
Yup, everything seems good. Thanks for contributing 😄

IlanCosman:
> But if you test a new icon and it looks good for you, I'm Ok with this changes.

Yah, this icon looks good with the recommended font, not sure about others but I'd like to keep it 🙂

> Rewrite "AWS_DEFAULT_*" env vars

There's no actual rewriting going on. The item uses `set -l` so all changes to aws variables are contained entirely within the function. Nothing changes for the user.

PR #330, opened 2022-08-18
Fix prompt when `$PATH` contains a space

### LINKED ISSUE #329: Recent update causes problems with items in $PATH containing spaces
Hi, a recent update of Tide or Fish shell (could be either one since I update them with the same script) causes an error on the Tide prompt wich has the result of the error being printed on opening of a new fish shell and on each new promt after finishing a command.

The error is apparently caused by the $PATH variable containing a path which has spaces in it. Temporarily removing this path from the $PATH variable makes the error dissappear. But since the path component is added by an external application on each fish shell initiation it's not possible to get it away save for installing the external application.
It looks like Tide is executing the components of the $PATH variable as commands which causes it to fail since the path component with spaces is not escaped.

Using latest version of fish shell with the default prompt does not cause the error.

The full error is:
```
fish: Unknown command: /Users/o/Snitch.app/Contents/Components
fish:
PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin:/Applications/Little Snitch.app/Contents/Components CMD_DURATION=4567 fish_bind_mode=insert set _tide_prompt_25948 (_tide_1_line_prompt)
                                                                                                                                           ^
```

9999years:
Recently, a directory containing a space got added to my `$PATH` variable. This breaks the `tide` prompt! The issue stems from incorrectly constructed `eval` strings; it's very difficult to get the quoting right, and hard to notice until you have a 'weird' character in a variable. Anyways, this patch fixes at least a few of these bugs.

#### Description

Fixes several bugs due to unsanitized `eval` inputs.

#### Motivation and Context

Closes #329 

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Prompt disappears entirely when `$PATH` contains a space (#329)
+
 ## [v5.4.0][] (Jul 29 2022)
 
 ### Features
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -37,8 +37,8 @@ if contains newline $_tide_left_items # two line prompt initialization
 function fish_prompt
     _tide_status=\$status _tide_pipestatus=\$pipestatus if not set -e _tide_repaint
         jobs -q && set -lx _tide_jobs
-        $fish_path -c \"set _tide_pipestatus \$_tide_pipestatus
-PATH=\$PATH CMD_DURATION=\$CMD_DURATION fish_bind_mode=\$fish_bind_mode set $prompt_var (_tide_2_line_prompt)\" &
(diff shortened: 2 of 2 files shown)


IlanCosman:
I think I found a simpler solution by just quoting `PATH`, but we'll see if it works. As for the other variables, it's not necessary to escape them since `CMD_DURATION` and `pipestatus` have a fixed and simple format. `fish_bind_mode` could be problematic, but it's not the case with the defaults, and no one should make it a problem by setting it to something with a space.

Thanks for working on this though, it's greatly appreciated :)

9999years:
Your solution will break as soon as there's a `$PATH` variable with a single quote in it. Fish offers a tool to solve this problem in the general case: `string escape`. Why not use it?

IlanCosman:
Yup, good catch.

PR #335, opened 2022-08-25
Add Pulumi support

devansh08:
#### Description

Add a right prompt section for displaying current Pulumi stack of the project. It reads the stack name from a local json file named  using hash of the project path. This section will only show up if the directory has a `Pulumi.yaml` file (similar to a `package.json` file for node), which is something that's always created by pulumi for a new project.

#### Motivation and Context

[Pulumi](https://www.pulumi.com/) is an IaaC tool that I use quite often from the command line and knowing the current stack that I am on before running commands is quite helpful.

Yellow prompt section in below screenshots. `dev` is the name of the current stack.

Lean
lean

Classic
classic

Rainbow
rainbow

#### How Has This Been Tested

#### Checklist

- [ ] I have updated the documentation accordingly. - Would like to know where and what do I need to update :)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -27,5 +27,5 @@ Closes # <!--- Please link to an open issue. -->
 <!-- Go over the following points and put an x in all the boxes that apply. -->
 <!-- If you're unsure about any of these, don't hesitate to ask! -->
 
-- [ ] I have updated the documentation accordingly.
+- [ ] I am ready to update the wiki accordingly.
 - [ ] I have updated the tests accordingly.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## [v5.5.1][] (Sep 14 2022)
+
+### Bug fixes
+
+- Fix errors when in weird directories like `has spaces` or `--has dashes` ([#346][])
+
+## [v5.5.0][] (Sep 5 2022)
+
+### Features
+
+- Add crystal language item ([#322][] - @rrevi)
+- Tide now looks in parent directories for special files, such as when displaying tool versions ([#323][])
+- Tide now truncates the beginning of git branch names instead of the end
+- A new variable, `tide_context_hostname_parts` lets you choose how many parts of the hostname, split on dots, to show. For example, if your hostname is `foo.bar.baz`:
+  - With `set tide_context_hostname_parts 5` you would see `user@foo.bar.baz`
+  - With `set tide_context_hostname_parts 1` you would see `user@foo` (this is the new default)
+  - With `set tide_context_hostname_parts 0` you would see `user`
+
+### Bug fixes
+
+- Display the correct versions of tools when PATH has been altered, e.g. by nvm ([#303][], [#330][] - @9999years)
+
 ## [v5.4.0][] (Jul 29 2022)
 
 ### Features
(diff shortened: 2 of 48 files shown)


IlanCosman:
Hi! Sorry for the delayed response, and thanks for contributing 🙂 

Unfortunately this isn't in a state I'd be comfortable merging. External commands like `grep` and `cat` should be replaced with fish builtins such as `string match` and redirection (`some_command <file.txt`). Those should be pretty simple to fix, but the biggest problem is with running python code, which is just not something I want to do.

Is there really no `pulumi print-stack` or `pulumi context` or something like that? That would be the optimal solution.

devansh08:
Hi, thanks for checking this out!

> External commands like grep and cat should be replaced with fish builtins such as string match and redirection (some_command <file.txt).

Will update those.

> Those should be pretty simple to fix, but the biggest problem is with running python code, which is just not something I want to do.
Is there really no pulumi print-stack or pulumi context or something like that? That would be the optimal solution.

I actually did find out couple of days ago there is a command `pulumi stack --show-name` which gives the stack name (I blame the poor `--help` description for not spotting this sooner :smile:). But it is a bit slow whenever I've used it, especially when a cloud service like AWS is used for storing the config/data. Honestly, I'm not sure why it takes so much time to just read a local file (current stack info is always stored in a local file, which is what my Python code reads).

For comparison, this is the time difference in using the command and using the Python script:
Screenshot_20220917_005617

It consistently takes ~1+ seconds. I think the speed of the pulumi command also depends on the number of stacks for the project, as I have seen this command take upto 3+ seconds on some actual projects, even while I was on a more powerful machine.

As a point, the method of reading the stack info is inspired from the Starship prompt, from [this function](https://github.com/starship/starship/blob/master/src/modules/pulumi.rs#L121) in particular. As noted in their comment as well, pulumi does not really have a command that's fast enough for prompt usage.

If there is a better fish/shell/anything based approach you (or anyone else here) can suggest for creating the SHA1 hash of the project path string, the Python dependency can be removed. Python was the fastest solution for calculating the hash I could come up with.

Thanks!

IlanCosman:
> If there is a better fish/shell/anything based approach you (or anyone else here) can suggest for creating the SHA1 hash

Indeed there is: `shasum` on MacOS (and bsd?) and `sha1sum` on Linux.

I'm on a mac right now, so the replacement for your python would be `echo -n $yaml_path | shasum | string sub -e40`. From benchmarking it is a 1.5x speedup. Afaict it should be the exact same code on Linux, just with `sha1sum`.

If you could remove the python code and instead check for `shasum` vs `sha1sum` (using `command -q`) and then use the appropriate one, that would be great :)

IlanCosman:
Aww snap don't tell anyone but just adding the `-S` flag to your python makes it faster than my `shasum` code. Still though I think it's better to use coreutils and 1. avoid necessitating python for this item, and 2. avoid `python` vs `python3` etc.

devansh08:
I actually forgot to mention in my previous message that I had tried `sha1sum` before going with Python and it did not work. But realized now I did not remove the newline from it :disappointed:. It does seem to be working correctly now with `sha1sum`. Will verify and update my code.

Thanks!

devansh08:
Hi, I've updated it with your suggestions. Lemme know your thoughts.

Thanks!

IlanCosman:
I made some improvements/changes to the code, feel free to take a look at them and comment/ask questions :)

devansh08:
Hey, the improvements look good to me! So many fish builtins I did not know about till this PR :smile:

Thanks!

IlanCosman (approved):

PR #336, opened 2022-08-25
Update to `rm` command in tests

devansh08:
#### Description/Motivation/Context/~Text Dump :P~

I was facing a common issue across multiple tests when I first tried to run `make test` in this repo. And being very new to the testing framework it took me quite a while to figure out what's wrong.

Basically I had my `rm` aliased (and forgotten about :) ) to `rm -v` which caused verbose output to stdout which then resulted as a failed case as the tests weren't expecting any extra lines of output. A simple change of all `rm`'s to `/bin/rm` fixed this for me.

IMO this type of change will bring about predictability and remove possible side effects (due to local aliases) with the test cases, which should be the case with test cases, as test cases should not be environment dependent. This could probably be extended to other core commands that are used in these tests as well.

Thoughts ?

#### How Has This Been Tested

Running `make test`

#### Checklist

### DIFF
--- a/tests/_tide_item_crystal.test.fish
+++ b/tests/_tide_item_crystal.test.fish
@@ -20,4 +20,4 @@ touch shard.yml
 
 _crystal # CHECK: ⬢ 1.5.0
 
-rm -r $crystal_directory
+/bin/rm -r $crystal_directory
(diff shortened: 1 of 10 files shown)


IlanCosman:
Good point, thanks for doing this :) However, the correct way to do it is to use [`command`](https://fishshell.com/docs/current/cmds/command.html). Could you change it to that?

devansh08:
Done!

IlanCosman:
Thanks for contributing 🙂

PR #341, opened 2022-09-01
Use prompt_hostname for obtaining hostname

jupblb:
#### Description
`prompt_hostname` is a function from fish-shell ([src](https://github.com/fish-shell/fish-shell/blob/master/share/functions/prompt_hostname.fish)). It is rather simple but useful when hostname is very long.

I have not tested this change.

### DIFF
--- a/functions/_tide_item_context.fish
+++ b/functions/_tide_item_context.fish
@@ -1,9 +1,9 @@
 function _tide_item_context
     if set -q SSH_TTY
-        tide_context_color=$tide_context_color_ssh _tide_print_item context $USER@$hostname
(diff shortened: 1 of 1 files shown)


IlanCosman:
I've written a more general solution with the new `tide_context_hostname_parts` variables. This lets you choose how many parts to show, so if your hostname is `foo.bar.baz`, and your `tide_context_hostname_parts` is `1`, then you would see `user@foo`. If your `tide_context_hostname_parts` is `0` then you would see `user`.

Thanks for trying to contribute though, I really appreciate it :)

PR #343, opened 2022-09-08
Display toolbox name

jazzsnobeatcake:
#### Description

Displays the toolbox name instead of the word toolbox.

#### Motivation and Context
$hostname defaults to "toolbox" when inside a toolbox.
$tide_toolbox_icon already indicates we're in a toolbox. 
Displaying the toolbox name adds further context in multi-toolbox environments.

#### How Has This Been Tested

I tested this on Fedora Silverblue 36

#### Checklist

Unsure if these are applicable.


daudix-UFO:
This would be really useful 👍

IlanCosman:
Hi, sorry for the delay on reviewing this. I haven't had access to a linux machine for a while, but will again soon.

daudix-UFO:
No problem ;)

IlanCosman:
Also, external commands like grep and cat should be replaced with fish builtins such as `string match` and redirection (`some_command <file.txt`). Makes the code much faster and more portable :)

PR #345, opened 2022-09-14
add proxy item

pendulm:
This PR adds the prompt item proxy, which checks and displays if any `_PROXY` env has been set.

#### Description
display an icon if any of `http_proxy` `all_proxy` `https_proxy` is presented (ignore case)

image

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_proxy.fish
+++ b/functions/_tide_item_proxy.fish
@@ -1,0 +1,5 @@
+function _tide_item_proxy
+    begin
+        env | grep -i -q 'all_proxy=\|http_proxy=\|https_proxy='
+    end && _tide_print_item proxy $tide_proxy_icon
+end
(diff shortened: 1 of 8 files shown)


IlanCosman:
Hi! Thanks for contributing :) Could you explain the use case for this item? Just poking around, I don't see any other prompt that have a proxy item (looking at powerlevel10k, starship, and spaceship).

pendulm:
It may be a personal case. 

In some network condition (like what I am in), you must use proxy to use some command(like brew, wget, git...) because of firewall or network interference.  So this item add a signal to tell you whether proxy is on.

deflock:
May be useful. But I have two questions:
1. Why did you choose this icon?
2. Does it make sense to show what proxy exactly you're using `䀹 http://127.0.0.1:1080`?

PR #365, opened 2022-12-25
Add Void Linux icon

Orthestral:
#### Description

Show Void Linux icon `f32e` in place of the default Linux icon if Void Linux is detected


IlanCosman:
Unfortunately the void icon is not available in the recommended font yet. 
It'll hopefully get patched in soon: https://github.com/romkatv/powerlevel10k/issues/2217

I'll be happy to merge your PR when that happens 🙂

PR #368, opened 2023-01-05
Add a note about git version

fastiuk:
Tide doesn't work with the git version less than 2.22

#### Description

Changed only the note in README file.

The reason why it doesn't work with lower git versions is because Tide uses `--show-current` option to determine the current branch, but that option was introduced in git v2.22

#### Motivation and Context

Notify the users about the requirements

#### How Has This Been Tested

#### Checklist


IlanCosman:
I don't generally notify users about minimum versions for items.

PR #381, opened 2023-02-17
Fix failing tests

primeapple:
#### Description

There were two failing tests.

The git one was because git needs a flag and an option to be set to add local repos as submodules.

The other one happened only locally on machines that have no english locale as default.

I fixed both of them :partying_face: 

#### Motivation and Context

Enable tests again ;-)

#### How Has This Been Tested

#### Checklist


IlanCosman (approved):

IlanCosman:
Great work, thanks for fixing these bugs, especially the git one, that would have been a bummer to debug 🙂

PR #390, opened 2023-03-22
FEAT: character icon for root

ftc2:
#### Description

add `tide_character_icon_root` (default: `#`)

pretty simple feature, not much to go wrong. tested on linux.
if you like it, i'm happy to update the wiki, other documentation, or whatever else you want. lmk.

image


IlanCosman:
Thanks for contributing to Tide :) 
Unfortunately, I don't think this looks very good, at least currently. And since it overlaps with another item (`context`), it has to look good. If there's a better icon than `#` then I'll reconsider.

PR #396, opened 2023-04-01
add gcloud

branchvincent:
#### Description

Adds a `gcloud` item to display the current project (like p10k). For performance, it reads the config files directly rather than invoking `gcloud config get-value project` (which takes ~350ms on my machine)

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_gcloud.fish
+++ b/functions/_tide_item_gcloud.fish
@@ -1,0 +1,8 @@
+function _tide_item_gcloud
+    set -q CLOUDSDK_CONFIG || set -l CLOUDSDK_CONFIG ~/.config/gcloud
+    path is $CLOUDSDK_CONFIG/active_config \
+        && read -l config <$CLOUDSDK_CONFIG/active_config \
+        && path is $CLOUDSDK_CONFIG/configurations/config_$config \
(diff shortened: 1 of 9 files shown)


functions/_tide_item_gcloud.fish
```diff
@@ -0,0 +1,8 @@
+function _tide_item_gcloud
+    set -q CLOUDSDK_CONFIG || set -l CLOUDSDK_CONFIG ~/.config/gcloud
+    path is $CLOUDSDK_CONFIG/active_config \
+        && read -l config <$CLOUDSDK_CONFIG/active_config \
+        && path is $CLOUDSDK_CONFIG/configurations/config_$config \
+        && string match -rg '^\s*project\s*=\s*(.*)' <$CLOUDSDK_CONFIG/configurations/config_$config | read -l project \
```
IlanCosman, line 6: We can use a named capture group here and avoid the `read`.

IlanCosman (changes requested):
Looks great overall, thanks for contributing :) Just the one small change and it should be ready to merge.

functions/_tide_item_gcloud.fish
branchvincent: good idea, thanks!

IlanCosman (approved):

PR #398, opened 2023-04-02
add direnv

branchvincent:
#### Description

Adds a [`direnv`](https://github.com/direnv/direnv) item to show its status, using the same color/icon as p10k.
The only addition is showing a different color when direnv will be active once the user allows it's execution.

#### Motivation and Context

As a brief background, direnv can be in three states for a given working directory:

1. Inactive (`.envrc` file does not exist)
1. Disallowed (`.envrc` exists but not yet executed as the user needs to explicitly allow it)
1. Active (`.envrc` exists and was executed)

This calls `direnv status` to distinguish between the last two states, which only takes ~12ms on my machine.
Note this won't be needed once https://github.com/direnv/direnv/pull/1010 has been merged.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_direnv.fish
+++ b/functions/_tide_item_direnv.fish
@@ -1,6 +1,6 @@
 function _tide_item_direnv
     set -q DIRENV_DIR || return
-    direnv status | string match -qr '^Found RC allowed false$' \
+    direnv status | string match -q 'Found RC allowed false' \
         && set -lx tide_direnv_color $tide_direnv_color_denied \
         && set -lx tide_direnv_bg_color $tide_direnv_bg_color_denied
     _tide_print_item direnv $tide_direnv_icon
--- a/functions/_tide_item_direnv.fish
+++ b/functions/_tide_item_direnv.fish
@@ -1,0 +1,7 @@
+function _tide_item_direnv
+    set -q DIRENV_DIR || return
+    direnv status | string match -qr '^Found RC allowed false$' \
+        && set -lx tide_direnv_color $tide_direnv_color_denied \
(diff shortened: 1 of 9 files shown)


functions/tide/configure/configs/classic_16color.fish
```diff
@@ -12,6 +12,11 @@ tide_context_color_root bryellow
 tide_context_color_ssh yellow
 tide_crystal_bg_color black
 tide_crystal_color brwhite
+tide_direnv_bg_color black
+tide_direnv_bg_color_denied black
+tide_direnv_color bryellow
+tide_direnv_color_denied brred
+tide_direnv_icon ▼
```
IlanCosman, line 19: Accidental copypaste into the 16color configs

functions/_tide_item_direnv.fish
```diff
@@ -0,0 +1,7 @@
+function _tide_item_direnv
+    set -q DIRENV_DIR || return
+    direnv status | string match -qr '^Found RC allowed false$' \
```
IlanCosman, line 3: No need for regex, the default behavior of `string match` is exact matching.

IlanCosman (changes requested):
Looks great. 2 small changes and it should be ready for merge. 

Thanks for contributing, I know making a new item can be a bit of a hassle with all the configs and all the different places you need to edit, but you have it down cold 😄

branchvincent:
fixed both, thanks for the quick review!

functions/tide/configure/configs/classic_16color.fish
branchvincent: oops

IlanCosman (approved):

PR #401, opened 2023-04-17
Feature: Truncate kubectl context using variables

rodrigobdz:
Truncate `kubectl` output using environment variables.

#### Description

Requires the following variables to be defined.

Example variable definition:
```sh
set -x tide_kubectl_truncation_strategy l
set -x tide_kubectl_truncation_length 31
```

#### Motivation and Context

My `kubectl` context is too large and almost fills an entire line.

Same approach as in:
https://github.com/IlanCosman/tide/blob/447945d2cff8f70d5c791dd4eec8b322d37798dd/functions/_tide_item_git.fish#L2

Closes # 

#### How Has This Been Tested

I tested the changes by modifying the file `$HOME/.config/fish/functions/_tide_item_kubectl.fish`.

#### Checklist

### DIFF
--- a/functions/_tide_item_kubectl.fish
+++ b/functions/_tide_item_kubectl.fish
@@ -1,4 +1,4 @@
 function _tide_item_kubectl
-    kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null | read -l context &&
+    kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null | string shorten -"$tide_kubectl_truncation_strategy"m"$tide_kubectl_truncation_length" | read -l context &&
         _tide_print_item kubectl $tide_kubectl_icon' ' (string replace -r '/(|default)$' '' $context)
 end


olujicz:
I can confirm that this is working on Linux.

rodrigobdz:
Thanks @olujicz. I checked the Linux test accordingly.

functions/_tide_item_kubectl.fish
```diff
@@ -1,4 +1,4 @@
 function _tide_item_kubectl
-    kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null | read -l context &&
+    kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null | string shorten -"$tide_kubectl_truncation_strategy"m"$tide_kubectl_truncation_length" | read -l context &&
```
0xMH, line 2: This doesn't work in MacOS 14.2.1.
Maybe use AWK instead of string? 
```
function _tide_item_kubectl
    set context (kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null | \
                 awk -v len="$tide_kubectl_truncation_length" '{print substr($0, 1, len)}')
    if set -q context[1]
        _tide_print_item kubectl $tide_kubectl_icon' ' (string replace -r '/(|default)$' '' $context)
    end
end
```

rodrigobdz: @0xMH I am open to using awk but the snippet above returns the beginning and not the end of the string.

Example:

The command

```sh
kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}'
```

returns the untruncated string:

```
arn:aws:eks:<region>:<numbers>:cluster/<clustername>/<namespace>
```

The snippet above returns `arn:aws:eks:<region>:<some_numbers>` instead of `<clustername>/<namespace>` for a given string length.

0xMH: Now? 
```
function _tide_item_kubectl
    set context (kubectl config view --minify --output 'jsonpath={.current-context}/{..namespace}' 2>/dev/null | \
                 awk -v len="$tide_kubectl_truncation_length" '{print substr($0, length($0)-len+1)}')
    if set -q context[1]
        _tide_print_item kubectl $tide_kubectl_icon' ' (string replace -r '/(|default)$' '' $context)
    end
end
```

rodrigobdz: Works

0xMH: what do you think @IlanCosman?

PR #415, opened 2023-06-15
Only remove unusable items if they were going to be used

brycekellogg:
#### Motivation and Context

On my setup (fish v3.6.1, Ubuntu 22.04, Windows 10 WSL2) the repeated calls to `type --query $cli_names` in `_tide_remove_unusable_items` causes noticeable lag when starting a new shell. Checking if the cli items are in the left or right prompt before doing the `type --query` speeds up `_tide_remove_unusable_items` for me by ~4.5x

#### How Has This Been Tested

I run fish with tide as my daily driver; I made the changes locally and ran with them for a couple days. I also ran `make test`, which passed.

#### Checklist

### DIFF
--- a/functions/_tide_remove_unusable_items.fish
+++ b/functions/_tide_remove_unusable_items.fish
@@ -1,7 +1,13 @@
 function _tide_remove_unusable_items
     # Remove tool-specific items for tools the machine doesn't have installed
     set -l removed_items
-    for item in aws chruby crystal direnv distrobox docker elixir gcloud git go java kubectl nix_shell node php pulumi rustc terraform toolbox virtual_env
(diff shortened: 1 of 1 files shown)


brycekellogg:
I don't currently have access to a MacOS system, so I won't be able to do any testing there in the near future. I also didn't see anywhere in the Wiki that needed to be updated related to this change.

The unit tests passed for me, but if you would like an additional test added that covers this change, let me know.

IlanCosman:
Thanks for the PR, I'll take a look at it soon :slightly_smiling_face: Don't worry about the mega-linter fail, it's buggy rn.

IlanCosman:
Done manually in dev branch, thanks for the change :)

PR #416, opened 2023-06-15
Display Python version in virtual_env item

bcyran:
#### Description

Display Python version alongside virtual env name.

#### Motivation and Context

When working on multiple Python projects I often have to deal with different Python versions. The version is just as important as the virtual env name.

#### How Has This Been Tested

Tested with a few virtual envs I have lying around, adjusted tests, ran `make test`.

#### Checklist


ilya-bobyr:
Seems like this has been already merged into `main` and slightly adjusted: https://github.com/IlanCosman/tide/commit/10a1fabc9ebdd3d2dd77bf0a23d6db411923b831

IlanCosman:
Yah, thanks for calling this out as an improvement @bcyran :)

PR #419, opened 2023-07-04
Add newline before prompt only when previous output exists

### LINKED ISSUE #327: Add newline before prompt only when previous output exists
First of all thanks for creating tide, I just tried fish and it's amazing to have a p10k equivalent for fish. This bug is just a minor annoyance but I don't know how to resolve it by myself :(!

#### Describe the bug

I stole the description [here](https://stackoverflow.com/questions/65722822/fish-shell-add-newline-before-prompt-only-when-previous-output-exists) because this is the exact same problem.

However, the newline is also printed when there is no previous output, e.g. after clearing the terminal with printf "\033c" (or when the terminal is first opened):

```
                           <--- bad newline: no previous output
➜ /some/dir            
❯ command1            
output...             
                           <--- good newline
➜ /some/dir          
❯ command2         
```

#### Steps to reproduce

Use `tide configure` and use prompt spacing `(2) Sparse` to see the issue.

#### Screenshots
This is what happens with tide and fish:
image

This is what happens with powerlevel10k and zsh:
image

Even the configuration wizard of tide advertises the behavior intended and not the reality:
image

#### Environment

```text
fish version: 3.5.1
tide version: 5.4.0
term: xterm-256color
os: macOS Monterey 12.5
terminal emulator: iTerm
fish startup: 11.29 millis
fisher plugins: jorgebucaran/fisher ilancosman/tide@v5
```

#### Additional context

I guess what we see is happening [here for double line prompt](https://github.com/IlanCosman/tide/blob/main/functions/fish_prompt.fish#L22) and [here for simple line prompt](https://github.com/IlanCosman/tide/blob/main/functions/fish_prompt.fish#L59) but I don't know how we can detect reliably we just started the interactive shell session or not.

CalaW:
#### Description

As #327 described, a newline is printed when there is no previous output. With this PR, the newline is omitted when the terminal is first opened or cleared using Ctrl-L.

Referred to the implementation from pure-fish, specifically in the files [_pure_prompt_new_line.fish](https://github.com/pure-fish/pure/blob/c0b83abec7197a0d06adc9ffe82d21d5a8c5f865/functions/_pure_prompt_new_line.fish#L6) and [_pure_init.fish](https://github.com/pure-fish/pure/blob/c0b83abec7197a0d06adc9ffe82d21d5a8c5f865/conf.d/_pure_init.fish#L5).

BTW, feel free to squash commits when merging!

#### Motivation and Context

Closes #327 

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/conf.d/_tide_init.fish
+++ b/conf.d/_tide_init.fish
@@ -1,6 +1,8 @@
 function _tide_init_install --on-event _tide_init_install
     set -U VIRTUAL_ENV_DISABLE_PROMPT true
 
+    set --global _tide_fresh_session true
+    
     source (functions --details _tide_sub_configure)
     _load_config lean
     _tide_finish
--- a/functions/fish_prompt.fish
+++ b/functions/fish_prompt.fish
@@ -20,7 +20,7 @@ function _tide_refresh_prompt --on-variable $prompt_var --on-variable COLUMNS
 end
 
 if contains newline $_tide_left_items # two line prompt initialization
-    test "$tide_prompt_add_newline_before" = true && set -l add_newline '\n'
+    test "$tide_prompt_add_newline_before" = true && test "$_tide_fresh_session" = false && set -l add_newline '\n'
 
     set_color $tide_prompt_color_frame_and_connection -b normal | read -l prompt_and_frame_color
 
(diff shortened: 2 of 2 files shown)


IlanCosman:
This does not work. The line you changed doesn't do what you think it does. The prompt is cached, so we'd need to check for the variable in every prompt run. I am still undecided if I want to do this. Thanks for trying to help tho :)

PR #425, opened 2023-07-27
Skip rebase/cherry-pick progress when it is absent

### LINKED ISSUE #378: Warning messages on "break" during an interactive git rebase
#### Describe the bug

The following warning messages appear on every new prompt draw:

warning: An error occurred while redirecting file '.git/rebase-merge/msgnum'
warning: Path '.git/rebase-merge/msgnum' does not exist
warning: An error occurred while redirecting file '.git/rebase-merge/end'
warning: Path '.git/rebase-merge/end' does not exist

#### Steps to reproduce

1. In a git repo start interactive rebase and set one commit to "break"
2. After returning to the prompt the warnings appear

My first analysis revealed, that the files `msgnum` and `end` don't exist on a commit "break", while on a commit "edit" they exist.

ilya-bobyr:
#### Description

Even when git is in rebase or cherry-pick mode, plan execution progress might be absent.
For example, when the plan is not prepared just yet - while `git rebase --interactive` is waiting for the editor to finish editing the rebase plan.

Here is a report of this issue: https://github.com/IlanCosman/tide/issues/378

Closes #378 

#### How Has This Been Tested

Only tested manually.  Not sure how to construct an automatic test for this case.
I do not have MacOS, but considering I am using commands that are already there, it should work on MacOS.

#### Checklist

  Does not seem to apply, as this is a bug fix.

  There are no automated tests that check rebase progress.
  For this particular case, it is even more nuanced, as we need to verify output while the interactive rebase command is waiting for the editor to exit.
  So, I just fixed the code.

  I only checked `git rebase`, but it does not seem to hurt to make a similar check for the `cherry-pick` case.

### DIFF
--- a/functions/_tide_item_git.fish
+++ b/functions/_tide_item_git.fish
@@ -14,16 +14,14 @@ function _tide_item_git
 
     # Operation
     if test -d $gdir/rebase-merge
-        # Turn ANY into ALL, via double negation.
-        if not path is --invert $gdir/rebase-merge/{msgnum,end}
+        # Turn ANY into ALL, via double negation
+        if not path is -v $gdir/rebase-merge/{msgnum,end}
             read -f step <$gdir/rebase-merge/msgnum
             read -f total_steps <$gdir/rebase-merge/end
         end
-        test -f $gdir/rebase-merge/interactive \
-            && set -f operation rebase-i || set -f operation rebase-m
+        test -f $gdir/rebase-merge/interactive && set -f operation rebase-i || set -f operation rebase-m
     else if test -d $gdir/rebase-apply
-        # Turn ANY into ALL, via double negation.
-        if not path is --invert $gdir/rebase-apply/{next,last}
+        if not path is -v $gdir/rebase-apply/{next,last}
             read -f step <$gdir/rebase-apply/next
             read -f total_steps <$gdir/rebase-apply/last
         end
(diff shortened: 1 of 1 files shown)


IlanCosman:
@ilya-bobyr Hey, thanks for fixing this 🙂 I made a small change for speed that shouldn't have any meaningful functional difference. Could you give it a test drive and see that it still works?

ilya-bobyr:
> @ilya-bobyr Hey, thanks for fixing this slightly_smiling_face I made a small change for speed that shouldn't have any meaningful functional difference. Could you give it a test drive and see that it still works?

Sorry for a slow reply.
A double negation to turn an "any" check into an "all" check is interesting :)
At first, I thought `not path is -v` is the same as `path is`.
I've added a comment, to clarify it is not a mistake.

Works for me.

It would be nice if fish would just provide `path is --all`.

IlanCosman:
> A double negation to turn an "any" check into an "all" check is interesting :)
> It would be nice if fish would just provide `path is --all`.

Agreed. I'll go submit an issue 😄

ilya-bobyr:
> > A double negation to turn an "any" check into an "all" check is interesting :)
> > It would be nice if fish would just provide `path is --all`.
> 
> Agreed. I'll go submit an issue smile

I briefly looked at the relevant `fish` source.  It is not completely trivial.
On top of that, it looks like `fish` is undergoing a rewrite in Rust.
So they probably would not want to accept PRs that change functionality.
An issue is probably the best!

PR #240, opened 2021-12-29
Add Ruby support

calebmeyer:
#### Description

Add a new prompt section (Tide Item) for Ruby. This should support package manager installs and Rbenv/RVM, though I've only tested with RVM

#### Screenshots (if appropriate)
image

#### How Has This Been Tested

- [ ] I have tested using **Linux**.
- [x] I have tested using **MacOS**.

image

#### Checklist

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #437, opened 2023-09-03
Converted chruby to generic ruby

cmer:
#### Description

Another Ruby implementation, following in the footsteps of #240.

I converted the chruby implementation to something that is more generic and supports any Ruby version manager such as chruby, rvm, asdf or even system installations.

CleanShot 2023-09-03 at 11 17 20@2x

#### How Has This Been Tested

#### Checklist


cmer:
This is another implementation of a Ruby item.

I did not adjust the tests because I'm not quite sure how to run them even. I could use some help on this front. Thank you!

IlanCosman:
Thanks for working on this! I refactored the code to be faster and smaller. It should hopefully have the same behavior 🙂 

A few thoughts:

#### 1

I don't love searching every directory above us for anything ending in `.gemspec`. It might get unreasonable on pathological filesystems, and anything that can happen will happen 😄 

I have absolutely no ruby-world knowledge, but according to my _very_ brief research (https://piotrmurach.com/articles/writing-a-ruby-gem-specification/), the convention is to name the file `package.gemspec`. Do you think just searching for that would work?

#### 2

Do we want to include `.tool-versions` in this search list? Is there any guarantee that the user is actually using the `.tool-versions` file to specify a ruby version, or could they be using it for something else and we're just showing useless information? This is the disadvantage of a general version manager I suppose.

halostatue:
> I have absolutely no ruby-world knowledge, but according to my _very_ brief research (https://piotrmurach.com/articles/writing-a-ruby-gem-specification/), the convention is to name the file `package.gemspec`. Do you think just searching for that would work?

It would not work. The convention is <code><em>package</em>.gemspec</code>. That is, I maintain a package called `diff-lcs`. Its gemspec is `diff-lcs.gemspec`.

> 
> #### 2
> Do we want to include `.tool-versions` in this search list? Is there any guarantee that the user is actually using the `.tool-versions` file to specify a ruby version, or could they be using it for something else and we're just showing useless information? This is the disadvantage of a general version manager I suppose.

We would not, IMO. If it `.tool-versions` is found, it would be better to inspect it to see if it contains `^ruby\s+\d`. Otherwise, `ruby --version` *might* error out.

functions/_tide_item_ruby.fish
```diff
@@ -0,0 +1,6 @@
+function _tide_item_ruby
+    if path is $_tide_parent_dirs/{*.gemspec,Gemfile,Rakefile,.ruby-version,.tool-versions}
```
halostatue, line 2: Might this be better to be something like

```fish
if command --query ruby && path is …
```

This would not help if we were using jruby or another version with a prefix, but it would prevent cases where ruby is selected but not yet installed.

halostatue (approved):

functions/_tide_item_ruby.fish
IlanCosman: That should be taken care of by this change in `_tide_remove_unusable_items`: 

https://github.com/IlanCosman/tide/pull/437/files#diff-fc249ac3efaa608703553322a4c2d675da3f1e4bd3456e861cdd638ca571409cR4

We check for `ruby` on startup.

cmer:
Thanks for the feedback, @IlanCosman . To answer your questions:

We actually do need to search for `*.gemspec` because gems typically don't use the name `package.gemspec`. For example, look at this gem that is part of Rails: https://github.com/rails/rails/tree/main/actioncable

The reason I was looking for `.tool-versions` was because I wanted to know which Ruby version is active in any directory I am in. But I understand others might not want that, so it's totally reasonable to remove it.

cmer:
Would it be overkill to also look for `*.rb`?

halostatue:
> The reason I was looking for `.tool-versions` was because I wanted to know which Ruby version is active in any directory I am in. But I understand others might not want that, so it's totally reasonable to remove it.

I can understand this, but I think that this is a viable request for pretty much any of the widgets. Something like `_tide_always ruby node` and checking for `contains -- ruby $_tide_always` in the Ruby plug-in would be a better approach for this sort of thing. Something like `$_tide_always` would be a different feature/pull request, though.

With respect to `*.rb`…I do think it would be overkill to search all parents; I would maybe search the current directory for `.rb` files, but even that is probably overkill.

cmer:
Cool let's merge this bad boy!

IlanCosman (approved):
Seems good to go, thanks for contributing :)

PR #445, opened 2023-09-25
Redirect stderr output of rustc --version

### LINKED ISSUE #444: rustup causes `rustc` invocation to automatically download toolchain
#### Describe the bug

If a rust folder contains rust-toolchain.toml requiring version that isn't installed, rustc --version automatically tries to install the toolchain. So if we quickly open nvim or any other cli program after switching to the folder, contents from stderr is displayed. 

#### Steps to reproduce

1. Create empty rust project with cargo init
2. Create rust-toolchain.toml file with content as follows:
`[toolchain]`
`# set channel to one that is not installed`
`channel = "1.72"`
4. Switch to the directory and open any CLI program like nvim or top.

#### Screenshots
image

arpandaze:
#### Description

Solves the problem described in #444 

#### Motivation and Context

Closes #444 
#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_rustc.fish
+++ b/functions/_tide_item_rustc.fish
@@ -1,6 +1,6 @@
 function _tide_item_rustc
     if path is $_tide_parent_dirs/Cargo.toml
-        rustc --version | string match -qr "(?<v>[\d.]+)"
+        rustc --version 2>/dev/null | string match -qr "(?<v>[\d.]+)"
         _tide_print_item rustc $tide_rustc_icon' ' $v
     end
 end


ilya-bobyr:
This would hide the installation process, but the toolchain will still be installed.
What the UX would be for the case when a toolchain for the current path is missing?
Installing the toolchain without any visual indication seems like a degradation, not an improvement.
And, I think, the prompt will also just hang until the installation is finished.

IlanCosman:
Yah, the prompt won't update until the download is finished, and even worse, the download will be restarted every time you press enter, because Tide kills the background shell that is doing the downloading.

Here's what Starship does with this same problem: https://github.com/starship/starship/blob/master/src/modules/rust.rs#L51-L89

> // `$CARGO_HOME/bin/rustc(.exe) --version` may attempt installing a rustup toolchain.
        // https://github.com/starship/starship/issues/417
        //
        // To display appropriate versions preventing `rustc` from downloading toolchains, we have to
        // check
        // 1. `$RUSTUP_TOOLCHAIN`
        // 2. The override list from ~/.rustup/settings.toml (like `rustup override list`)
        // 3. `rust-toolchain` or `rust-toolchain.toml` in `.` or parent directories
        // 4. The `default_toolchain` from ~/.rustup/settings.toml (like `rustup default`)
        // 5. `rustup default` (in addition to the above, this also looks at global fallback config files)
        // as `rustup` does.
        // https://github.com/rust-lang/rustup.rs/tree/eb694fcada7becc5d9d160bf7c623abe84f8971d#override-precedence
        //
        // Probably we have no other way to know whether any toolchain override is specified for the
        // current directory. The following commands also cause toolchain installations.
        // - `rustup show`
        // - `rustup show active-toolchain`
        // - `rustup which`

We should probably copy their logic. I'd be happy to accept PRs, but I probably won't be working on this myself in the near term.

PR #469, opened 2023-12-14
Add sudo prompt item

re-xyr:
#### Description

Adds a simple prompt item that displays an icon if `sudo` can be called without a password. 

#### Motivation and Context

This is most useful in reminding the user that `sudo` has been called recently and hasn't timed out yet.

#### How Has This Been Tested

It correctly displays an icon when `sudo` can be called without a password, and displays nothing otherwise.

#### Checklist


IlanCosman:
I don't really get why this would be useful. Could you describe a situation where this would be helpful?

IlanCosman:
To be honest, this seems like security theater, or at the very least encouraging really bad practices. I don't like it, sorry :(

PR #533, opened 2024-10-06
Bug Fix Array Element Removal in _tide_pwd.fish

psycoder-sup:
#### Description

This pull request addresses an error occurring in the Fish shell prompt function _tide_pwd.fish. The original line of code attempts to remove an element from the glob array using an incorrect approach involving the contains command. When the list is empty this code shows error:

```shell
~/.config/fish/functions/_tide_pwd.fish (line 23):
            set -e glob[(contains -i $parent_dir/$dir_section/ $glob)] # This is faster than inverse string match
            ^
in function '_tide_pwd'
in command substitution
in command substitution
in function '__fish_prompt_orig'
	called on line 47 of file -
in function 'fish_prompt'
in command substitution
```

So I fixed this line of code to handle empty list situation.

#### Motivation and Context

Closes # 

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_pwd.fish
+++ b/functions/_tide_pwd.fish
@@ -28,8 +28,10 @@ eval "function _tide_pwd
             string match -qr \"(?<trunc>\..|.)\" \$dir_section
 
             set -l glob \$parent_dir/\$trunc*/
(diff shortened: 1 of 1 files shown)


yanshay:
This issue happened to me as well and this PR seem to have solved it.

PR #535, opened 2024-10-16
Which terraform

### LINKED ISSUE #534: Unknown command terraform
#### Describe the bug

When I access the directory of a `conda` or `pixi` environment which is not activated by default and that environment uses `terraform`, I get the following error because `terraform` is not available:

```
❯ fish: Unknown command: terraform                                                                                                                                           
~/.config/fish/functions/_tide_item_terraform.fish (line 3):                                                                                                                 
        terraform workspace show | string match -v default | read -l w &&                                                                                                    
        ^~~~~~~~^                                                                                                                                                            
in function '_tide_item_terraform'                                                                                                                                           
        called on line 15 of file ~/.config/fish/functions/_tide_2_line_prompt.fish 
```

#### Steps to reproduce

1. Create a `test` directory with a `.terraform` subdirectory: `mkdir -p /tmp/test/.terraform`
2. Change to the `test` directory: `cd /tmp/test`
3. See error from description

#### Environment

Output of `tide bug-report`: 

```text
fish version: 3.7.0
tide version: 6.1.1
term: screen-256color
os: 
terminal emulator: gnome-terminal
fish startup: 376.04 millis
fisher plugins: jorgebucaran/fisher meaningful-ooo/sponge ilancosman/tide@v6 jorgebucaran/nvm.fish jethrokuan/z franciscolourenco/done patrickf1/fzf.fish
```

jcollado:
#### Description

Check that `terraform` binary is available before trying to use it.

#### Motivation and Context

Remove error messages when a `.terraform` directory is present, but the `terraform` binary is not available.

Closes #534

#### How Has This Been Tested

I have used the code in this PR and the error was gone.

#### Checklist


jcollado:
A couple of comments regarding the checklist:
- I believe that the wiki doesn't need to be updated.
- I've updated the test cases so that they pass. However, I don't know what would be the right way to add a test to verify the situation when the terraform command is not found.

PR #553, opened 2024-12-22
Fix an error when Python is not installed

knyazer:
#### Description

I'm using a dev-container setup with distrobox, with everything besides the OS shared between the container and the host system. Python is not installed on the host system, but it is installed in the container. All vars are shared between the system and the container, so the host thinks that it is in the python environment, while the environment is not installed. 

This fix shows nothing if there is no python available.

#### Motivation and Context

Shows error if VIRTUAL_ENV is well-defined, but python is not there.

Closes # 

#### How Has This Been Tested

Just checked on my machine; the change is trivial, so I don't think it needs heavy testing..

#### Checklist

- [x] I am ready to update the wiki accordingly. (not relevant)

### DIFF
--- a/functions/_tide_item_python.fish
+++ b/functions/_tide_item_python.fish
@@ -2,9 +2,11 @@ function _tide_item_python
     if test -n "$VIRTUAL_ENV"
         if command -q python3
             python3 --version | string match -qr "(?<v>[\d.]+)"
-        else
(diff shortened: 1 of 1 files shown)


knyazer:
Oh, yikes, I just copied the code instead of formatting it well... will finish it in a few hours :)

Issue #564, opened 2025-03-03
Some items are not working after fish 4.0 update


augustocdias:
I have noticed the items that rely on check if specific files are in the current tree are not working. (for example rust and node). Basically the `path is $_tide_parent_dirs` is not working correctly.

Debugging it, I've noticed the `path is ` is fine, so I believe the generation of the env variable is not being done correctly. We're getting a `''` entry in there.

nnungest:
Maybe something like this:

```fish 
        string escape (
            for dir in (string split / -- $PWD)
                if test (string length $dir) -lt 1
                    continue
                end
                set -la parts $dir
                string join / "" $parts
            end
        )
```

nnungest:
I took a stab at this in https://github.com/IlanCosman/tide/pull/566

PR #566, opened 2025-03-06
fix: restore path based parsing in 4.0

nnungest:
This fixes issue 564, where empty directories are returned after upgrading to 4.0.0. This also safely wraps paths.

#### Description

Restore path based parsing for a variety of _tide_item functions. 

#### Motivation and Context

When a user changes directory into a one that contains, lets say package.json, tide normally picks up the version and prints it on the status line. This stopped working when fish moved to 4.0.0. This restores the previous behavior by checking for directories that are invalid (zero length) and safely wrapping the directories with quotes.

Closes https://github.com/IlanCosman/tide/issues/564

#### How Has This Been Tested

I ran this against the make commands list in contributing.

- [X] I have tested using **Linux**. [fish 3.7.1] [nixos]
- [X] I have tested using **MacOS**. [fish 3.7.1] [fish 4.0.0] [MacOS Sequoia 15.3.1]

#### Checklist

- [X] I am ready to update the wiki accordingly. **no changes needed**
- [X] I have updated the tests accordingly. **no changes needed**

### DIFF
--- a/functions/_tide_parent_dirs.fish
+++ b/functions/_tide_parent_dirs.fish
@@ -1,7 +1,11 @@
 function _tide_parent_dirs --on-variable PWD
-    set -g _tide_parent_dirs (string escape (
-        for dir in (string split / -- $PWD)
-            set -la parts $dir
-            string join / -- $parts
-        end))
+    set -g _tide_parent_dirs (
+        string escape (
+            for dir in (string split / -- $PWD)
+                if test (string length \"$dir\") -lt 1
+                    continue
+                end
+                set -a parts $dir
+                string join "" (string join / --  $parts) ""
+            end))
 end
--- a/functions/_tide_parent_dirs.fish
+++ b/functions/_tide_parent_dirs.fish
@@ -2,11 +2,8 @@ function _tide_parent_dirs --on-variable PWD
     set -g _tide_parent_dirs (
         string escape (
             for dir in (string split / -- $PWD)
-                if test (string length \"$dir\") -lt 1
-                    continue
-                end
                 set -fa parts $dir
-                string join "" (string join / --  $parts) ""
+                string join / -- $parts
             end
         )
     )
--- a/functions/_tide_parent_dirs.fish
+++ b/functions/_tide_parent_dirs.fish
@@ -5,7 +5,7 @@ function _tide_parent_dirs --on-variable PWD
                 if test (string length \"$dir\") -lt 1
                     continue
                 end
-                set -a parts $dir
+                set -fa parts $dir
                 string join "" (string join / --  $parts) ""
             end))
 end
--- a/functions/_tide_parent_dirs.fish
+++ b/functions/_tide_parent_dirs.fish
@@ -7,5 +7,7 @@ function _tide_parent_dirs --on-variable PWD
                 end
                 set -fa parts $dir
                 string join "" (string join / --  $parts) ""
-            end))
+            end
+        )
+    )
 end


nnungest:
Im not sure if the last 2 checks apply to this PR but im happy to update things if needed. 

In this case I don't think the test should be touched as this should keep the same  functionality as before intact.

0xflick:
Thanks! this fixed my issue.

Looks like @IlanCosman hasn't been active much recently, but hopefully this makes its way in

d0sse:
Thanks!

functions/_tide_parent_dirs.fish
```diff
@@ -1,7 +1,11 @@
 function _tide_parent_dirs --on-variable PWD
-    set -g _tide_parent_dirs (string escape (
-        for dir in (string split / -- $PWD)
-            set -la parts $dir
-            string join / -- $parts
-        end))
+    set -g _tide_parent_dirs (
+        string escape (
+            for dir in (string split / -- $PWD)
```
faho, line 4: This is `string split --no-empty`

functions/_tide_parent_dirs.fish
```diff
@@ -1,7 +1,11 @@
...
+                if test (string length \"$dir\") -lt 1
+                    continue
+                end
+                set -a parts $dir
```
faho, line 8: This will override a global or universal $parts. The variable should be defined as local outside of the loop, or use `set -fa`.

I'm not sure why this ever worked, tbh.

faho (changes requested):

functions/_tide_parent_dirs.fish
nnungest: ok ive resolved this part by adding the `f` flag to set within that loop

functions/_tide_parent_dirs.fish
nnungest: various incantations of  `--no-empty` did not work. ill try again soon.

nnungest-aca: @faho would you be able to explain a bit more perhaps where --no-empty would go or what would be replaced by it? Thank you in advance!

nnungest: Thank you for responding in so much detail. I'll clean this up this weekend and run it through the tests again.

scaryrawr: Testing locally for myself, the main change that appears to be needed is the `-la` to `-fa`.

With no change I see each folder listed out individually:

```fish
echo $_tide_parent_dirs
'' Users mike GitHub tide
```

With `-fa` I get the full paths, plus an empty string:

```fish
echo $_tide_parent_dirs
'' /Users /Users/mike /Users/mike/GitHub /Users/mike/GitHub/tide
```

If I use the `--no-empty` with `-fa` it's missing the leading `/`, so the paths are invalid:

```fish
echo $_tide_parent_dirs
Users Users/mike Users/mike/GitHub Users/mike/GitHub/tide
```

I can add the leading slash

```fish
function _tide_parent_dirs --on-variable PWD
    set -g _tide_parent_dirs (string escape (
        for dir in (string split --no-empty / -- $PWD)
            set -fa parts $dir
            echo /(string join / -- $parts)
        end))
end
```

And then I get:

```fish
echo $_tide_parent_dirs
/Users /Users/mike /Users/mike/GitHub /Users/mike/GitHub/tide
```

But I think it may be simpler to just switch to `-fa` from `-la` and it would be closer to the fish-3 behavior :thinking:?

And should we have just `/` ? I don't expect to find anything at root to trigger something that uses `$_tide_parent_dirs`, but just wondering.

scaryrawr: The code here:

```fish
if test (string length \"$dir\") -lt 1
  continue
end
```

Is escaping the quotes, so it'll never be less than 1. It'll always at least be 2, so we're never hitting the `continue`

The string join at the end:

```fish
string join "" (string join / -- $parts) ""
```

Is using empty strings to join the output to empty string.

So I think they can be removed?

nnungest: did you run the tests? "make test"

nnungest: for whatever reason removing `\"$dir\"` and using just  `(string length $dir)` makes the tests fail.

nnungest: similarly, changing `-fa` to `-la` causes the test to fail as well

scaryrawr: I have:

```fish
function _tide_parent_dirs --on-variable PWD
    set -g _tide_parent_dirs (string escape (
        for dir in (string split / -- $PWD)
            set -fa parts $dir
            string join / -- $parts
        end))
end
```

And the `tests/_tide_parent_dirs.test.fish` passes:

```
Testing file tests/_tide_parent_dirs.test.fish ... ok (89 ms)
```

nnungest: ah that looks great as well. let me push this up

nnungest: @faho  I think this can be marked resolved now

nnungest-aca: this should was resolved in https://github.com/IlanCosman/tide/pull/566/commits/16d8dba77dbec2e53d22e9228c710c7c3c43240f

nextzhou:
Thank you. Hope to merge soon！

augustocdias:
It seems the owner is not very active anymore. Is anyone willing to start a fork?

nnungest-aca:
> It seems the owner is not very active anymore. Is anyone willing to start a fork?

ive since started using starship instead.

nickmonad:
@plttn Thank you! I started using that fork. I setup a new machine and was going crazy over missing icons. I couldn't understand where the failure was coming from until I realized the new machine is on fish 4 and the old machine was on fish 3 and I was hitting that parent dirs issue

PR #570, opened 2025-03-15
Tweak status item; adds cmd_duration_precise item

jetpks:
##### Tweaks status item
- Adds a new configuration option `tide_status_simple_failures` containing status codes that should be considered "simple failures"
  - Default value added to `configure/configs/*`
  - Tests added
- I plan to set this empty so all failures show a right-side indicator
- This lets you have both the character color changing and the indicator

##### Adds cmd_duration_precise item
- Almost the same as cmd_duration, but with this behavior:
  - when duration is > 1h, only show hours & minutes e.g. "1h 23m"
  - when duration is < 1h && > 1m, show minutes & seconds (no decimals) e.g. "1m 23s"
  - when duration is < 1m && > 1s, show seconds with decimals e.g. "1.23s"
  - when duration is < 1s, show milliseconds e.g. "123ms"
- Continues to obey tide_cmd_duration_threshold and tide_cmd_duration_decimals
- Tested

#### Motivation and Context
For the status item, I always want a failure indicator on the right side _and_ I want the character to change color (I don't always notice the color change, but it's nice to have.)

For cmd_duration_precise, I have two aims:
1. I like to see timings for all commands, but I don't want to see output like "0.427s" and would prefer "427ms"
2. I don't need to see the smaller time units when the total duration is quite large. e.g. when a command takes 9h 47m, i don't really care if it also 23s or 53s of that 47th minute. cmd_duration_precise makes all outputs ~5-ish chars long depending on the size of the numbers while still allowing you to see large and miniscule durations.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_cmd_duration_precise.fish
+++ b/functions/_tide_item_cmd_duration_precise.fish
@@ -1,0 +1,17 @@
+function _tide_item_cmd_duration_precise
+    test $CMD_DURATION -lt $tide_cmd_duration_threshold
+    and return 0
+
+    set d $CMD_DURATION
+    if test $d -ge 3600000
+        set out (math -s0 $d / 3600000)'h '(math -s0 $d / 60000 % 60)'m'
+    else if test $d -ge 60000
+        set out (math -s0 $d / 60000 % 60)'m '(math -s0 $d / 1000 % 60)'s'
+    else if test $d -ge 1000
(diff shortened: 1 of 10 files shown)


jetpks:
will reopen when ready; originally meant to open this against my fork lol

PR #573, opened 2025-03-28
Add support for `nix shell` detection

### LINKED ISSUE #542: Detect `nix shell` (in addition to `nix-shell`)
The nix_shell item currently detects whether it is inside a nix shell by looking at `IN_NIX_SHELL`, but that var [is not set when using the new nix command's `nix shell`](https://github.com/NixOS/nix/issues/3862).

Some discussion in that issue talks about searching for nix store entries in `PATH` ([which is summarized here](https://discourse.nixos.org/t/in-nix-shell-env-variable-in-nix-shell-versus-nix-shell/15933/2)).

I think Tide should adopt a similar strategy for detecting `nix shell` usage as well.

Prior art:
- https://github.com/starship/starship/blob/fccf4ccebf1533989234ec13c5a0f6d44e47e937/src/modules/nix_shell.rs#L32-L40

scaryrawr:
#### Description

When using `nix shell`, the `IN_NIX_SHELL` variable is not set.

A common workaround is to check if `/nix/store/` is on the PATH.

#### Motivation and Context

`nix shell` does not set `$IN_NIX_SHELL`

Closes #542

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_nix_shell.fish
+++ b/functions/_tide_item_nix_shell.fish
@@ -1,3 +1,13 @@
 function _tide_item_nix_shell
-    set -q IN_NIX_SHELL && _tide_print_item nix_shell $tide_nix_shell_icon' ' $IN_NIX_SHELL
+    if set -q IN_NIX_SHELL
+        _tide_print_item nix_shell $tide_nix_shell_icon' ' $IN_NIX_SHELL
+    else
+        for path_item in $PATH
+            if string match -q '/nix/store/*' $path_item
+                # if nix is ran with `nix shell` we cannot determine pure/impure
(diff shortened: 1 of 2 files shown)


scaryrawr:
> > A common workaround is to check if /nix/store/ is on the PATH.
> 
> Is that really a common workaround? Wouldn't that always identify a NixOS environment as a nix shell?

Ah, maybe not common, but it is how [starship](https://github.com/starship/starship/blob/80cc1ba2f2e35e23aa821e4a84c8aba5e4b18efc/src/modules/nix_shell.rs#L15-L30) treats it. They have an `unknown` option vs `pure`/`impure` where if they're in an NixOS environment.

PR #575, opened 2025-03-30
Update _tide_item_rustc.fish

### LINKED ISSUE #444: rustup causes `rustc` invocation to automatically download toolchain

road2react:
Fixes #444 by using `rustup show active-toolchain` instead of running `rustc`, which is downloaded on-demand.

### DIFF
--- a/functions/_tide_item_rustc.fish
+++ b/functions/_tide_item_rustc.fish
@@ -1,6 +1,6 @@
 function _tide_item_rustc
     if path is $_tide_parent_dirs/Cargo.toml
-        rustc --version | string match -qr "(?<v>[\d.]+)"
+        rustup show active-toolchain | string match -qr "(?<v>^[\d.]+)"
         _tide_print_item rustc $tide_rustc_icon' ' $v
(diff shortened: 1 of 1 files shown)


ErrorNoInternet:
Wouldn't this no longer show the version? It just shows an icon for me.

```sh
$ rustup show active-toolchain
nightly-x86_64-unknown-linux-gnu (default)
```

antoyo:
Is there anything that prevents this from being merged?
Thanks for your work on this!

PR #595, opened 2025-08-10
fix: preserve color formatting when setting 'normal' background

laermannjan:
#### Description

From the fish docs
> Using the normal keyword will reset foreground, background, and all formatting back to default.

As `tide_"$item"_bg_color` is often set to "normal", it currently resets any formatting (e.g. `--bold`) that applied to the `tide_"$item"_color`.
Setting the background color separately and prior to the foreground color allows for this formatting to be preserved.

Before
image

After:
image

### DIFF
--- a/functions/_tide_print_item.fish
+++ b/functions/_tide_print_item.fish
@@ -14,7 +14,8 @@ function _tide_print_item -a item
         echo -ns $tide_right_prompt_separator_diff_color
     end
 
-    v=tide_"$item"_color set_color $$v -b $item_bg_color
+    set_color -b $item_bg_color
+    v=tide_"$item"_color set_color $$v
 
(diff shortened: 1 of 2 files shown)


laermannjan:
A problem I just noticed is that in many cases tide doesn't reset the formatting before setting a new color.
For example at [here](https://github.com/IlanCosman/tide/blob/main/functions/_tide_item_git.fish#L64-L71).
So a `--bold` for one of the earlier segments (e.g. stash) would also be applied to later ones.

This is probably out of scope for this PR (and maybe you don't want to support this kind of formatting at all) .
I guess we could add a `apply_color` wrapper where we wrap coloring an item into `set_color normal` calls.
Would be happy to supply another PR or add to this one, but wanted to get your thoughts on this first.

PR #616, opened 2025-12-21
use fish-native transient prompt

branchv:
This is now supported in fish since 4.2.1: https://github.com/fish-shell/fish-shell/pull/11153


plttn:
I've added this in my fork as of (my) v7.0.0 FYI. Works great along with other patches I've collected from here.

PR #626, opened 2026-01-07
Handle empty fish_key_bindings in fish 4.3 (fixes #620, #622)

### LINKED ISSUE #620: Incorrect character icon in fish 4.3.1
#### Describe the bug

After upgrading to fish 4.3.1, the wrong icon is chosen for the character item with the default keybindings.

#### Steps to reproduce

1. Update to fish 4.3.1
2. Use the default key bindings and default character icon
3. See that the prompt character is the vi mode default icon

#### Screenshots

```text
❮ fish --version
fish, version 4.3.1
❮ echo "$fish_bind_mode"
default
❮ echo "$fish_key_bindings"
fish_default_key_bindings
❮ echo $tide_character_icon
❯
❮ echo $tide_character_vi_icon_default
❮
```

#### Environment

Output of `tide bug-report`: 

```text
fish version: 4.3.1
tide version: 6.2.0
term: tmux-256color
os: MacOS Tahoe 26.1
terminal emulator: iTerm2
fish startup: 26.82 millis
fisher plugins: ilancosman/tide@v6 jorgebucaran/fisher
```

#### Additional context

It appears as though `$fish_key_bindings` is unset while executing `_tide_item_character` for some reason.  It is set in an otherwise empty `fish_prompt` function.

### LINKED ISSUE #622: Backwards Prompt Icon
On MacOS with Fish installed from homebrew. I did an update and Fish was updated. The prompt icon (greater than sign) is now backwards. It's a less than sign. How do I get it back?

Image

In the `tide configure` setup, it shows correctly

Image

sstieger:
Fish 4.3 seems to leave `$fish_key_bindings` empty by default, leading the check for default bindings in `_tide_item_character` to fail, and causing tide to render the vi-mode icons instead of the standard prompt character.

#### Description

Updated `_tide_item_character` to include a check for empty `$fish_key_bindings`.

Closes #620.
Closes #622.

#### How Has This Been Tested

#### Checklist

### DIFF
--- a/functions/_tide_item_character.fish
+++ b/functions/_tide_item_character.fish
@@ -3,7 +3,7 @@ function _tide_item_character
 
     set -q add_prefix || echo -ns ' '
 
-    test "$fish_key_bindings" = fish_default_key_bindings && echo -ns $tide_character_icon ||
+    test -n "$fish_key_bindings" && test "$fish_key_bindings" != fish_default_key_bindings &&
         switch $fish_bind_mode
             case insert
                 echo -ns $tide_character_icon
@@ -13,5 +13,6 @@ function _tide_item_character
                 echo -ns $tide_character_vi_icon_replace
             case visual
                 echo -ns $tide_character_vi_icon_visual
-        end
+        end ||
+        echo -ns $tide_character_icon
 end


sstieger:
Oh, missed #619 when creating this, since it wasn't mentioned in the two issues.

MalcolmWardlaw:
It's probably a little pedantic, but doesn't this style of running test with && and || a little non-robust. Might it be better to make it a formal if/else thing?

PR #649, opened 2026-04-01
Add vbattery and hbattery prompt items

### LINKED ISSUE #294: Batteries (not) included
#### Is your feature request related to a problem? Please describe

Many of my devices (smartphone, laptops) run on battery, it would be great if `tide` includes battery item out of the box.

#### Describe the solution you'd like

I was thinking…
```fish
function _tide_item_battery
  if which termux-battery-status &>/dev/null
    set -l termux_bat_out (termux-battery-status)
    set bat_capacity (string match -rg 'percentage": ([0-9]*)' $termux_bat_out)
    set bat_status (string match -rg '"status": "([^"]*)"' $termux_bat_out)
  else if test -d /sys/class/power_supply/BAT0
    set bat_capacity (cat /sys/class/power_supply/BAT0/capacity)
    set bat_status (cat /sys/class/power_supply/BAT0/status)
  else
    return
  end

  switch (string lower $bat_status)
    case charging
      set bat_status CHG
    case discharging
      set bat_status DIS
    case *
      set bat_status '??'
  end
  if [ $bat_capacity -lt 15 ]
    set_color red
  else if [ $bat_capacity -lt 30 ]
    set_color yellow
  else
    set_color green
  end
  printf "%s%%" $bat_capacity
  set_color normal
  printf " (%s)" $bat_status
end

```

When added to right prompt items it results in:
obrázok

jose1711:
Two new items for displaying battery status:
  - `vbattery`: font icons with 11 charge levels, separate charging/discharging icon sets and color alerts for low/critical
  - `hbattery`: block bar with 5 levels using integer bucketing

Both items share common configuration under `tide_battery_*` (method, thresholds, critical/low colors) while each has its own `tide_<item>_bg_color` and `tide_<item>_color`.

Supported backends: upower (default), sysfs, termux.

#### Motivation and Context

Closes #294

obrázok

#### How Has This Been Tested

Added both items to left prompt items on a laptop running on battery. Verified visually. Tested via `littlecheck.py`.

#### Checklist

### DIFF
--- a/functions/_tide_item_hbattery.fish
+++ b/functions/_tide_item_hbattery.fish
@@ -1,0 +1,49 @@
+function _tide_item_hbattery
+    switch $tide_battery_method
+        case termux
+            set -l termux_bat_out (termux-battery-status)
+            set -f bat_capacity (string match -rg 'percentage": ([0-9]*)' $termux_bat_out)
+            set -f bat_status (string match -rg '"status": "([^"]*)"' $termux_bat_out)
+        case sysfs
+            if set -q tide_battery_sysfs_name
+                set -f bat_capacity (cat /sys/class/power_supply/$tide_battery_sysfs_name/capacity)
+                set -f bat_status (cat /sys/class/power_supply/$tide_battery_sysfs_name/status)
+            else
+                set -l battery_name /sys/class/power_supply/BAT*
+                if count $battery_name >/dev/null
+                    set -f bat_capacity (cat $battery_name/capacity)[1]
+                    set -f bat_status (cat $battery_name/status)[1]
+                end
+            end
+        case '*'
+            set -l upower_out (upower -b)
+            set -f bat_capacity (string match -rg 'percentage: *([0-9]*)' $upower_out)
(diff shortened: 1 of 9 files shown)


jose1711:
This also extends #578 (however lacks percentage which could be added separately)

Martian-0007:
Perhaps fish could do a toggle in a variable for the percentage display?
If I understood the code correctly, you're already polling for it anyway

Either way, thanks a lot for the work

jose1711:
> Perhaps fish could do a toggle in a variable for the percentage display? If I understood the code correctly, you're already polling for it anyway
> 
> Either way, thanks a lot for the work

Thank you. Instead of a toggle I was thinking of adding a separate item (`pbattery`?). This way user can choose whether he/she wants to see just the icon, numeric value or both.

Martian-0007:
Wouldn't a separate item also add another separator? I like the combination of icon and value, as it helps with orientation at a glance.

iirc p10k also uses a little clock for time (maybe I just haven't tampered with the config of tide enough to find that yet)

image

PR #661, opened 2026-05-26
Go item: detect go.work for Go workspace support

### LINKED ISSUE #660: Go version indicator should support Go workspaces: detection of `go.work`
#### Is your feature request related to a problem? Please describe

Currently, the `go` item extension in Tide only detects if a user is inside a Go project by checking for the existence of a `go.mod` file. However, Go has natively supported multi-module workspaces via `go.work` files since Go 1.18.

If a developer opens a shell in the root of a Go workspace that contains a `go.work` file but no top-level `go.mod` file, Tide fails to display the Go prompt item and version indicator.

#### Describe the solution you'd like

The `go` prompt item should become visible if either a `go.mod` **or** a `go.work` file is present in the current directory or any parent directories.

- Current Implementation (`functions/_tide_item_go.fish`):

  ```fish
  function _tide_item_go
      if path is $_tide_parent_dirs/go.mod
          go version | string match -qr "(?<v>[\d.]+)"
          _tide_print_item go $tide_go_icon' ' $v
      end
  end
  ```

- Proposed solution:

  ```fish
  function _tide_item_go
      if path is $_tide_parent_dirs/go.mod $_tide_parent_dirs/go.work
          go version | string match -qr "(?<v>[\d.]+)"
          _tide_print_item go $tide_go_icon' ' $v
      end
  end
  ```

#### Steps to Reproduce

1. Create a new directory and initialize a go workspace: `mkdir go-workspace && cd go-workspace && go work init`
2. The Tide prompt does not display the Go version indicator.

#### Environment

```text
fish version: 4.7.1
tide version: 6.2.0
term: xterm-kitty
os: Arch Linux x86_64
terminal emulator: kitty 0.46.2
fish startup: 12.08 millis
fisher plugins: ilancosman/tide@v6
```

llawn:
Currently the Go prompt item only checks for go.mod files. Go has
supported multi-module workspaces via go.work files since Go 1.18.

Add go.work to the path check so the Go version indicator is shown
in Go workspace roots that don't have a top-level go.mod.

Closes #660

### DIFF
--- a/functions/_tide_item_go.fish
+++ b/functions/_tide_item_go.fish
@@ -1,5 +1,5 @@
 function _tide_item_go
-    if path is $_tide_parent_dirs/go.mod
+    if path is $_tide_parent_dirs/go.mod $_tide_parent_dirs/go.work
         go version | string match -qr "(?<v>[\d.]+)"
         _tide_print_item go $tide_go_icon' ' $v
     end
--- a/functions/tide/configure/configs/classic.fish
+++ b/functions/tide/configure/configs/classic.fish
@@ -79,7 +79,7 @@ tide_pwd_bg_color 444444
 tide_pwd_color_anchors $_tide_color_light_blue
 tide_pwd_color_dirs $_tide_color_dark_blue
 tide_pwd_color_truncated_dirs 8787AF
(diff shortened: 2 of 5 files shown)


llawn:
I have also updated `_tide_item_go.test.fish` to include test coverage for this change, and appended `go.work` to the `tide_pwd_markers` list inside `functions/tide/configure/configs/{classic,lean,rainbow}.fish`. Now all files after `grep -rI "go.mod"` have been modified.

PR #669, opened 2026-07-27
fix(pwd): guard truncation-branch string ops with -- for leading-dash dirs

### LINKED ISSUE #668: Prompt breaks (and can hang) in dash-leading directories when path truncates
# Prompt breaks (and can hang) in directories whose name starts with `-` when the path is long enough to truncate

## Description

When the current directory contains a path component that starts with a dash (`-`) **and** the path is long enough to trigger PWD truncation, `_tide_pwd` errors:

```
string join: -foo: unknown option
test: Missing argument at index 3
math: Error: Too few arguments 'max 0,'
```

If a same-prefix sibling directory also exists (so the truncation logic has to grow the abbreviated prefix), it's worse: `string escape --style=regex $trunc` errors once `$trunc` becomes e.g. `-a`, the disambiguation `while` loop never advances, and **the prompt hangs**.

A short dash-leading path (e.g. `~/tmp/--has dashes/foo`) is fine, because it never enters the truncation branch — which is exactly why the existing test for dash dirs doesn't catch this.

## Root cause

In `functions/_tide_pwd.fish`, the truncation branch passes path-derived values to `string match`, `string escape`, and `string join` without a `--` separator, so a leading-dash value is read as an option. The non-truncating path already guards its `string join` calls with `--`; the truncation branch does not. Same class of bug as fish-shell/fish-shell#10169 (`prompt_pwd`).

## Reproduction

```fish
set d /tmp/dashbug/-a-very-long-leading-dash-directory-name-aaaaaaaaaaaaaaaaaaaa/bbbbbbbbbbbbbbbbbbbbbbbbb/ccccccccccccccccc/d
mkdir -p -- $d
cd $d          # prompt renders the error cascade

# hang variant: add a same-prefix sibling so truncation must disambiguate
mkdir -p -- /tmp/dashbug/-a-very-long-leading-dash-directory-name-aaaaaaaaaaaaaaaaaaab
cd $d          # prompt hangs
```

## Environment

- tide: current `main`
- fish: (fill in `fish --version`)
- OS: macOS (reproduced); OS-independent — pure fish string-option parsing

## Fix

Add `--` before the path-derived argument in the five affected calls in the truncation branch. PR to follow.

Frefreak:
#### Description

`_tide_pwd`'s truncation branch passes path-derived values to `string match`, `string escape`, and `string join` without a `--` separator. When a directory component starts with `-`, fish parses it as an option:

- `string join` / `string match` → `unknown option`, the length read fails, and the prompt collapses into a cascade of `test: Missing argument` / `math: Too few arguments` errors;
- worse, `string escape --style=regex $trunc` inside the disambiguation `while` loop errors once `$trunc` grows to e.g. `-a`, so the loop never advances and **hangs the prompt** when a same-prefix sibling exists.

The non-truncating path already guards its two `string join` calls with `--`; the five uses inside the truncation branch were missed. This adds `--` before the path-derived argument in each.

#### Motivation and Context

Directories starting with `-` are legal and occur in practice (git worktrees, tools that encode an absolute path as a slug, etc.). Any such dir deep enough to trigger PWD truncation breaks the prompt; with a same-prefix sibling it hangs the shell. Same class of bug as fish-shell/fish-shell#10169 in `prompt_pwd`.

Closes #668

#### How Has This Been Tested

Added regression cases to `tests/_tide_item_pwd.test.fish` covering a leading-dash directory under truncation, both without and with a same-prefix sibling (the hang case). The pre-existing `--has dashes` case only covered a short path that never truncates, which is why this went unnoticed.

#### Checklist


M1nt-Ch0c0:
### Automated Linux verification

I ran a targeted regression check for this PR on:

- Arch Linux
- fish 4.8.1
- PR commit `1c7b384`

Results:

- Current `main` reproduces `string join: -alfa: unknown option` followed by the `test: Missing argument` cascade in a sufficiently long path containing a leading-dash component.
- This PR passes both targeted cases: truncation with `-alfa`, and disambiguation with a same-prefix sibling (`-alfawolf`). Both completed with status 0 and without errors or a hang.

I did not run the full test suite; this was a focused verification of the new regression scenarios. Linux side looks good for this fix.

— **GPT-5.6-Sol** *(automated agent)*

mbolli:
Confirming this on Linux (the PR checklist only has macOS ticked).

Fedora 44, fish 4.6.0, tide `main` @ fcda500, running `tests/_tide_item_pwd.test.fish` through littlecheck in an isolated `XDG_CONFIG_HOME`:

1. **Pristine `main` + its own test file: green.** Confirms the point in #668 that the existing `--has dashes` case never enters the truncation branch, so the bug goes uncaught.
2. **New tests + unpatched `_tide_pwd.fish`: fails as described.** `string join: -alfa: unknown option`, then the `test: Missing argument at index 3` cascade on every redraw.
3. **This PR: green** (243 ms).

One correction on the hang, in case it matters for the issue text: I could *not* reproduce a hang on pristine `main` at fish 4.6.0, with either the repro in #668 or a real-world four-sibling dash tree. Unpatched, the loop's second `string match ... $dir_section` also errors (rc 2), which terminates the `while`, so you get the error cascade but the prompt still returns.

The hang is real, but it needs the *first* `string match` guarded while the `string escape --style=regex $trunc` calls are not: `$trunc` resolves to `-`, escape errors to an empty pattern, and `(?<trunc>.)` re-matches the same first character forever. That is the state you land in if you apply only part of this fix, which reads as an argument for landing all five `--` together rather than against the PR.

Hit this via Claude Code scratchpad dirs, which slugify an absolute path into a leading-dash directory name (`/tmp/claude-1000/-home-user-project/...`), so it is easy to reach without doing anything exotic.

schlomo:
Must have for Claude users... Can you please merge this so that we don't need to maintain a fork?

